PR #4, opened 2016-03-05
Remove TinyXML dependency

nderwin:
This pull request removes the TinyXML dependency that was declared in the MML project.  There are no code changes, just changes to build scripts.

This also includes a couple of bug fixes found while making changes and testing

- the DOS batch file was still using Java 1.6, so this was updated to 1.7
- the packaging_local.xml file referenced pkgdir when it wasn't set (localdir was set instead)


Dylan-M:
Please don't do multiple things in a single pull request. Especially since you've reverted the start of some work and I can't refuse it without refusing the entire pull request or pulling it manually on the command line and modifying it.

You reverted the setting to localdir, when the change to pkgdir was correct and just needed to be completed in the file. I had started it and apparently forgot to finish it, but pkgdir is the correct variable.

nderwin:
Well, that's fine I can back that change out.  I just noticed that it was broken while I was testing.

arlith:
I'm going to merge this anyways, as the TinyXML bit is needed so MML can build with the trunk of MM.  Just try to keep PRs to a single "topic" in the future.

PR #8, opened 2016-03-23
Print pilot names

vizax:
Updated Coords for Pilot Name/Skills on:
- Mech
- Quad
- Tripod

### DIFF
--- a/docs/history.txt
+++ b/docs/history.txt
@@ -1,6 +1,7 @@
 VERSION HISTORY:
 ----------------
 v0.1.52-git
++ PR #8: Adjust position of printed pilot names for Meks
 
 v0.1.51 (2016-02-21 14:15 UTC)
 + Updated MegaMek.jar to 0.41.15
--- a/src/megameklab/com/ui/Aero/Printing/PrintAero.java
+++ b/src/megameklab/com/ui/Aero/Printing/PrintAero.java
@@ -35,6 +35,8 @@ import javax.print.attribute.HashPrintRequestAttributeSet;
 import javax.print.attribute.standard.PrintQuality;
 
 import megamek.common.Aero;
+// TODO: uncomment when print issue is fixed and pilot data is ready to position
+// import megamek.common.Crew;
 import megamek.common.TechConstants;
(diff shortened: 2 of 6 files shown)


vizax:
I checked the rest too.
`PrintAero.java` and `PrintSmallCraftSpheroid.java` had commented out the pilot fetching code. In my local files I have the code cleaned up and theoretically ready to go, but since I wasn't 100% positive, I didn't want to add them to this PR. Other things were missing too, like the `import megamek.common.crew` call. I'd like to work on these more, but I wanted to get some good data submitted while I take some time for the rest of it.

comments/suggestions/thoughts welcome. :-)

arlith:
Multi-crew cockpits for 'mechs is still a pending feature for MM, so I don't know that you'll be able to get printing information for the gunner.  The rest of it looks good and straightforward.

arlith:
One other thing is, there's a docs/history.txt.  When you make a bug fix or add a feature, it's good to add a line to that file.  For instance, for this one you could add "+ PR #8: Adjust position of printed pilot names for Meks".  This file gets used as the change log when releases are made, so try to word things so they would make sense to users.

vizax:
Thanks for the info. I'll be sure to update the history file for this and the Indy TSM update I did too.

I did have problems last night with the Aero testing. I'm putting in static strings for reference and then just running MML through eclipse, and all the rest has worked fine, except Aero. When I print to (either to PDF or an actual printer) it hangs on the actual print. Example: I see the PDF conversion window open, and the progress bar may or may not start, but then it just sits. I've let it sit for up to 15 minutes.  Thoughts on this? better/different way?

arlith:
Have you tried checking the logfile to see if there were any errors in the log?  It's possible there's a bug with Aero printing.

vizax:
Silly me. However, there's nothing in the log. When I step through the code, it hangs on the below (see screenshot).
debug clip

I wasn't thinking straight last night, so I tested the actual release for MML, and cannot print an Aero sheet from it either. I'm going to table this for now, and raise the non-printing as a separate issue.

last commit is added code for when the print issue is fixed.

Issue #23, opened 2016-08-27
Several Issue with the Mac Version


LeftoverNoodles:
MLL is not working well on my Mac.  I have a with the Java6SE and the JRE8 installed.  The few times things have outright crashed it's been while running Java8.

Here is a quick rundown 

- App version won't launch / generate a log file.
- The raw Jar versions won't launch via a command line.
- When launching the raw Jar file via finder is locks up when loading a cache (http://imgur.com/a/hIlyi)
- When launching the Jar file from within the app package is crashes when you try to print (JRE8 error)

**Launching the App with the Open -n command**

LSOpenURLsWithRole() failed with error -10810 for the file /Users/Geg/Documents/Battletech/MegaMek/megameklab-0.1.57-mac/MegaMekLab.app.

**When trying to load the Jar with Java6:**

Exception in thread "main" java.lang.UnsupportedClassVersionError: megameklab/com/MegaMekLab : Unsupported major.minor version 51.0
	at java.lang.ClassLoader.defineClass1(Native Method)
	at java.lang.ClassLoader.defineClassCond(ClassLoader.java:637)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:621)
	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:141)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:283)
	at java.net.URLClassLoader.access$000(URLClassLoader.java:58)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:197)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:190)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:306)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:301)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:247)
Friday:Commands Geg$ 

**When Trying to Load the raw Jar from the command line (JRE8):**

Unable to redirect output
Memory Allocated [3817865]
java.io.FileNotFoundException: ./data/fonts/Eurosti.TTF (No such file or directory)
	at java.io.FileInputStream.open0(Native Method)
	at java.io.FileInputStream.open(FileInputStream.java:195)
	at java.io.FileInputStream.<init>(FileInputStream.java:138)
	at megameklab.com.util.UnitUtil.loadFonts(UnitUtil.java:1640)
[...]

PR #24, opened 2016-08-29
Fix for printing issue in issue #23.

wildj79:
This will fix the printing issues on OS X reported in issue #23. There was an issue where trying to print from OS X was causing MML to crash. The problem seems to be that the Graphics2D.drawImage() calls in printMekImage() in both the PrintMech and PrintTripod classes were implemented incorrectly.  They included a parameter for color which was causing issues.  Updating these methods to use the correct method call fixes the issue.

### DIFF
--- a/src/megameklab/com/ui/Mek/Printing/PrintMech.java
+++ b/src/megameklab/com/ui/Mek/Printing/PrintMech.java
@@ -886,7 +886,7 @@ public class PrintMech implements Printable {
         int height = Math.min(200, img.getHeight(null));
         int drawingX = 237 + ((148 - width) / 2);
         int drawingY = 172 + ((200 - height) / 2);
-        g2d.drawImage(img, drawingX + leftMargin, topMargin + drawingY, width, height, Color.BLACK, null);
+        g2d.drawImage(img, drawingX + leftMargin, topMargin + drawingY, width, height, null);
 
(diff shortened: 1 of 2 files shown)


arlith:
Out of curiosity, how did you figure that out?  Do you know why the color parameter caused issues?

wildj79:
@arlith I started testing other types of units to see if they had the same issue.  When I tried printing a Quad design, it worked.  So I started looking at the differences between the two. The difference that stuck out was the `Graphics2D.drawImage` calls in `printMekImage` in both classes. Which, when I stepped through with the debugger is where the crash would occur.  Then I started googling how to use the `drawImage` method, which led me to the conclusion that the call in the `PrintMech.printMekImage` method was incorrect.  As to why the color parameter was causing issues, I have no idea.  I'm not a Java developer by trade, so my knowledge of the frameworks used by Java are very limited. But, my google-fu is pretty good 😉

beerockxs:
The Color parameter was there on purpose. Because we earlier draw some SVGs where the lastg color used is not necessarily black, we change it at this point.

Dylan-M:
@wildj79 Tried to find your email on your GitHub page, it wasn't there. So... dropping you a message here that you should join our public slack team.

http://megamek.org:3000/

PR #41, opened 2016-11-27
Added naval movement modes.

neoancient:
That was surprisingly easy. I haven't tested it much, but it loads naval units from the cache. Printing is broken.


arlith:
Without reading the rules and digging into it more, I can't really validate these changes.  The only thing I could say is to check the rules for the construction of naval units and make sure we're following those rules.  If that checks out, then it's probably good.  Additionally, we try to keep all of the "rules adherence" portion within classes in `megamek.common.verifier`, particularly for this `TestEntity` and `TestTank`.  You may have to update some of these classes to verify certain aspects of naval units.

neoancient:
All the necessary code was already present in `Tank` and `TestTank`. All I had to do was add the options in MML. It took more time to double-check the verification than to make the changes.

PR #54, opened 2017-03-13
Added the ability to select XCT specialization

pheonixstorm:
This should complete the addition of XCT troops into Megamek.

### DIFF
--- a/src/megameklab/com/ui/Infantry/views/SpecializationView.java
+++ b/src/megameklab/com/ui/Infantry/views/SpecializationView.java
@@ -85,13 +85,13 @@ public class SpecializationView extends IView implements ActionListener {
             
             gbc.gridx++;
             gbc.anchor = GridBagConstraints.CENTER;
-            if (spec == Infantry.TAG_TROOPS || spec == Infantry.PARATROOPS) {
+            if (spec == Infantry.TAG_TROOPS || spec == Infantry.PARATROOPS || spec == Infantry.XCT) {
                 specPanel.add(new JLabel("-"), gbc);
             } else {
                 specPanel.add(new JLabel("10"), gbc);
             }
             gbc.gridx++;
-            if (spec == Infantry.TAG_TROOPS || spec == Infantry.PARATROOPS) {
+            if (spec == Infantry.TAG_TROOPS || spec == Infantry.PARATROOPS || spec == Infantry.XCT) {
                 specPanel.add(new JLabel("-"), gbc);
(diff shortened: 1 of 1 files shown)


HammerGS:
This is likely not ready to go. Since XCT needs sub-specialization.  Per this thread.

https://github.com/MegaMek/megamek/pull/493#issuecomment-286023060

I'm still waiting to hear back from Xotl for additional clarifications.

Dylan-M:
Additionally, we need to know (if the thread doesn't already specify) whether or not infantry can be trained for multiple specializations. I would hope yes, and honestly I want this written for the yes answer regardless so that it's possible to override with an option to allow it if the rules say no.

pheonixstorm:
From a coding standpoint I still think splitting XCT up isn't needed. You can't call a snowsuited XCT rated for Vacuum and expect them to survive. Doing so would require adding A LOT of XCT specs.

All
Cold Weather
Extreme Heat
Tainted Atmosphere
Toxic Atmosphere
Vacuum.
Hazardous Liquids

I can see wanting to get the rules right, but sometimes you just have to fudge things a bit to make it work. Its not done often, but it is done to make things work. Anyone else agree?

Dylan-M:
No, I don't. I didn't say send snowsuited infantry that are also trained for vacuum operations into space in their snowsuits. If they can have multiple specializations, it also stands to ready they would have equipment for each specialization. If we ever start tracking infantry equipment, that will have to be dealt with. But, that should _never_ be a deterrent from doing it _right_. That attitude is why the code is such a cluster f... mess right now. I will not perpetuate that.

pheonixstorm:
There is no equipment, it is all armor based. And that is something you can't change w/o putting the unit back in MML to swap out the armor from one kit to another

Dylan-M:
Unless we make a provision from doing it via the context menu.
Change Infantry Armor -> Snowsuit
                                          Vac Suit
                                          etc...

This is easily do-able.

pheonixstorm:
And save over the old blk or create a new one? I see where your thinking is but not sure how best to work the end result. Would MM take something like this from HQ or would you have to save/reload?

Dylan-M:
It would be no different than how we currently track ammo for 'Mechs. Nothing gets saved to the actual unit file, it's tracked in the campaign file as a modification to the unit there.

pheonixstorm:
Which code wise I have never looked into... Ok, will add all the specs to MM/MML.

PR #81, opened 2017-09-21
Armor allocation view

neoancient:
This continues the work I started when I rewrote the structure tab to integrate the IO tech advancement data. Since the armor allocation section did not deal with tech advancement I left it for a separate change.

Instead of separate classes for each type of unit that allocates armor by location (mech, vehicle, fighter at this point), there is one class to manage the armor values for a single location and another that creates and positions the location panels and shows the summary data.

I have also removed the patchwork dialog and replaced it with a view below the allocation panel as a persistent indicator of what type of armor is installed in each location and an easier way to change the armor for a location. There are also fixes for the current armor weight in the summary panel and status bar.

One other thing I did was consolidate the listeners for the various individual structure tab subpanel into a set of more general listener interfaces.

I intend to hold this until after the next development snapshot. I also realized as I was preparing this that I forgot to add the auto-allocate button to the new armor view so this is not quite ready anyway.


neoancient:
Auto-armor allocation done, with some fixes in mechs to spread better and vehicles to put the reduced rear armor in the rear of superheavies instead of the right rear.

PR #111, opened 2017-12-05
Update NC3 FCS tonnage display

mkerensky:
Per the Naval C3 construction rules on TO p332, any FCS tonnage for having more than (12) weapons on a given facing is doubled if a Naval C3 is installed. This is in addition to the 1% of the ship base weight of the system.  This adds the mml side to DropshipCriticalView.java


neoancient:
This only changes what is displayed on one tab in MML. It doesn't have any effect on the weight calculations used for validation.

mkerensky:
I know. There is a second PR in MM that handles the weight. This one requires it to work correctly.

neoancient:
The extra tonnage should be applied where it's calculated rather than having to find all the places that use the value and doubling it there.

PR #114, opened 2017-12-08
Record sheet printing overhaul, part 1

neoancient:
Overhaul of mech printing using the id attribute to identify elements within the SVG document and modify them in place, rendering the entire document to the print system's Graphics2D object when finished. This removes the need to have the positions of various pieces of data hard-coded and theoretically allows alternate and user-customized record sheets. It also improves the final quality. This approach was @arlith's idea and I first tried it out with conventional infantry.

What has changed:
1. All mech record sheets are handled by a single class, since the data is the same (with the exception of additional movement modes for LAMs and QuadVees, which are handled conditionally).
2. There is a hierarchy of printing classes, with `PrintMech` extending `PrintEntity`, which extends `PrintRecordSheet`. As more units types are reworked there is no need to repeat code. Having a separate `PrintRecordSheet` base class allows for future implementation of non-Entity record sheets, such as handheld weapons or advanced buildings.
3. Armor and structure pip printing is handled by an algorithm that distributes fairly evenly. The current method of having a set layout for each value offers more control over design, but is labor-intensive and limited, and is also tied to having a specific record sheet design. The pip algorithm uses svg <rect> elements to define rows on the armor or structure diagram at the default pip size, but can resize the pips if there is insufficient space, removing the need to determine an upper limit. This is especially necessary for vehicles and aerospace units, which have a maximum armor value for the unit rather than location, and have an absolute armor value by location that far exceeds anything typical (or reasonable).
4. Added fields for quirks, SPAs (when printing from a mul), and unit role.

What this fixes:
1. Printing for superheavies and tripods (and superheavy tripods).
2. Show extra crew data position for command console as well as tripod, superheavy tripod, and QuadVee cockpits.
3.Unit-specific sheets for LAMs and QuadVees.

Here's a representative sample of the output (18 pages).
https://drive.google.com/open?id=1nhfCBBRrZoSoqw_rXebmt-FHxpNIiBUb


neoancient:
Here's a representative sample of the output (18 pages).
https://drive.google.com/open?id=1nhfCBBRrZoSoqw_rXebmt-FHxpNIiBUb

beerockxs:
Some small issues:
- quadvee sheets have two copyright characters at the very bottom
- quad internal structure diagram has underlines under some of the IS values, which is in the wrong place for left and right torsos
- the "Overflow" text at the top of the heat scale does not look right for quads and LAMs
- the omniquadvee and quadvee damage transfer diagram is missing the arrows
- the font for the location names in the armor and IS diagrams looks too thin compared to the original sheets, was it bold there?
- the heat data for LAMs has some issues with letters being in the wrong place
- it'd be nice if tonnage, tech base, rules level and role lined up
- in the right torso location, the crit text for LRM 15 Artemis V Ammo is too big, overlapping with the border
- in the superheavy sheets, the location names in the armor and IS diagrams overlap with the dropshadow of the armor/IS diagram
- for three-legged mechs, "Engine Hits", "Gyro Hits", "Sensor Hits" and "Life Support" are not bold
- all pilots with abilities have them print twice

neoancient:
While looking at the overlap between the armor/is text and drop shadows I noticed that there are a few places where the position of the armor/is value is not quite right, so I'll adjust those as well the next chance I get to work on it.

The text for armor and structure is not bold in the original sheets, but I could experiment with making it bold and seeing how it looks.

I don't see any underlining in the quad internal structure diagram, though it does have some of the asymmetric text positioning I mentioned above. I also don't see any misalignment of the tonnage, tech level, rules level, and role. This is what I'm seeing there:

image
image

beerockxs:
Maybe it's the spacing between the letters for the text for armor and structure? I think having it like the text for non-hittable crits would be good.

Here's how the Goliath IS diagram looks in my Acrobat Reader:
grafik

I meant lining up the left edge of the 55, Inner Sphere, Standard and Brawler

neoancient:
Finished fixing those issues and uploaded a new version of the examples. I also added two units to display features I had forgotten about. The Malice shows armored components and modular armor, and also indicates an armor type that affects game play (in this case Ferro-Lamellor). The Zeus-X shows patchwork armor that affects game play (reactive) but only in some locations (arms and legs).

beerockxs:
When a 'Mech has single heat sinks, the number of heatsinks is right-aligned instead of left, and thus not directly above the heat sink pips. I think having it aligned with the heat sink pips is better and correct.

neoancient:
So would you have the number left-aligned even with the row of pips or align the pips with the right side of the region?

beerockxs:
I'd do the first.

PR #134, opened 2018-01-20
Line Ending, white space, permission fixes.

rjhancock:


rjhancock:
Adding the execute bit as they were failing on OS X.

PR #159, opened 2018-06-06
Use generic chassis image on RS if specific model not found.

neoancient:
If no fluff image is found for the model, check for generic chassis image for record sheet.

Issue reported here: https://bg.battletech.com/forums/index.php?topic=61783.msg1418023#msg1418023

### DIFF
--- a/src/megameklab/com/util/ImageHelper.java
+++ b/src/megameklab/com/util/ImageHelper.java
@@ -228,21 +228,18 @@ public class ImageHelper {
         }
 
         path = new File(path, dir).getAbsolutePath();
-        f = new File(path, unit.getShortNameRaw() + ".png");
-        if (f.exists()) {
-            return f;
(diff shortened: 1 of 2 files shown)


src/megameklab/com/util/ImageHelper.java
```diff
@@ -228,21 +228,18 @@ public static File getFluffFile(Entity unit, String dir) {
...
-            return f;
-        }
-        f = new File(path, unit.getShortNameRaw() + ".jpg");
-        if (f.exists()) {
-            return f;
-        }
-        f = new File(path, unit.getShortNameRaw() + ".jpeg");
-        if (f.exists()) {
-            return f;
+        final String [] EXTENSIONS = { ".png", ".PNG", ".jpg", ".JPG", ".jpeg", ".JPEG", ".gif", ".GIF" };
```
NickAragua: It's good enough for now. Ideally, there'd be some mechanism to map a specific unit type or chassis to a picture, but that's way out of scope for this.

Issue #178, opened 2018-08-08
Equipment Table getting stuck on one equipment type.


UB-JM:
### Environment
MegaMekLab Version: 0.44.0
-Downloaded from github, alone, on the day this issue was created. 
OS: Windows 10 Home
Java: Java 8, Build 1.8.0_181-b13. 

### Description
Switching the selected equipment type in the Equipment window may bug the tables. 

I wanted to check if there is any weapon dealing 14 damage. 
With a fresh download, I stuck with the preselected mech, chose "Experimental" and "Mixed Inner Sphere", then switched to Equipment, chose "All Weapons", and sorted by damage. The result was that weapons were sorted by type, then by damage, and that wouldn't change anymore. I then tried just a specific weapon type, and I get the issue at the latest after the third switch. I have not been able to reproduce the issue without sorting by damage. 
Afterwards, only a restart helps - Whatever weapon type I'm stuck at, I can't get out of that selection. All other Tabs still work normally, though. 
Resizing may have happened at first usage.
![snipping](https://user-images.githubusercontent.com/19310656/43837257-3c81c0de-9b18-11e8-8859-b5a23a0319db.PNG)

I will try to reproduce this on a different PC running Windows 10 N Professional on Sunday. 
(Assuming this is the right place to report this)

ndrwstn:
I've had that problem using MacOS. I believe I end-run it by checking/unchecking "show unavailable" as that seems to force a refresh of the table. A similar problems occurs when you use the filter box too much.

wildj79:
As always, logs would be useful.

UB-JM:
[megameklablog.txt](https://github.com/MegaMek/megameklab/files/2270858/megameklablog.txt)
Like this? 

I've now retried and shortest I can do is starting the program, going to equipment, and filtering by damage. At the latest after 2 switches (energy -> ballistic -> missile) it bugs out.

wildj79:
Thanks, the logs were exceptionally helpful here. Looks like the sorter is crapping out on weapons that have AoE damage. Those weapon types have an `A` after the damage value, which insn't being parsed correctly. Should be a relatively easy fix.

sixlettervariables:
Closed by #179

PR #179, opened 2018-08-08
Fix for sorting by damage in equipment tables

neoancient:
The table sorter converts the string value for weapon damage to an int value (floating point for infantry) but does not account for all the available formats of the damage string. It can handle a numeric value or a set of damages by range separated by a slash.

This change truncates the string when the first non-digit value is reached before parsing. If there is no leading digit, it treats it as damage 0.

### DIFF
--- a/src/megameklab/com/ui/BattleArmor/tabs/EquipmentTab.java
+++ b/src/megameklab/com/ui/BattleArmor/tabs/EquipmentTab.java
@@ -789,20 +789,30 @@ public class EquipmentTab extends ITab implements ActionListener {
                 return -1;
             } else {
                 //get the numbers associated with each string
-                float r1 = parseDamage(s1);
-                float r0 = parseDamage(s0);
-                return ((Comparable<Float>)r1).compareTo(r0);
+                double r1 = parseDamage(s1);
+                double r0 = parseDamage(s0);
+                return ((Comparable<Double>)r1).compareTo(r0);
             }
         }
 
-        private float parseDamage(String s) {
-            float damage = 0;
-            if(s.contains("/")) {
-                damage = Float.parseFloat(s.split("/")[0]);
-            } else {
-                damage = Float.parseFloat(s);
+        /**
+         * Extracts a numeric value from the damage string for use in sorting by damage. If a String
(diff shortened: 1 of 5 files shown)


neoancient:
Fix for #178

src/megameklab/com/ui/BattleArmor/tabs/EquipmentTab.java
```diff
@@ -789,20 +789,30 @@ public int compare(String s0, String s1) {
...
+         * @param s A weapon damage string
+         * @return  The value to use for sorting.
+         */
+        private double parseDamage(String s) {
+            s = s.replaceAll("[^0-9\\.]+", "/");
+            if (s.contains("/")) {
+                s = s.substring(0, s.indexOf("/"));
+            }
+            if (s.length() > 0) {
+                return Double.parseDouble(s);
```
sixlettervariables: This would be more likely to choke than the `Integer.parseInt` above, would it be better to `try...catch` with a warning log and return 0? Would keep the UI trucking along and the log would still be preserved.

src/megameklab/com/ui/Infantry/views/WeaponView.java
```diff
@@ -417,10 +417,24 @@ public int compare(String s0, String s1) {
...
         private double parseDamage(String s) {
-            double damage = 0;
-            damage = Double.parseDouble(s);
-            return damage;
+            s = s.replaceAll("[^0-9\\.]+", "/");
+            if (s.contains("/")) {
+                s = s.substring(0, s.indexOf("/"));
+            }
+            if (s.length() > 0) {
+                return Double.parseDouble(s);
```
sixlettervariables: Same as above.

src/megameklab/com/ui/BattleArmor/tabs/EquipmentTab.java
neoancient: There's no need to log anything. There are perfectly valid input strings that do not contain any numeric value, such as "Special". What this does is check for those and return zero for sorting purposes. Any other value is truncated at the first non-digit character before parsing.

PR #188, opened 2018-08-22
PrintMech: Weapons & Equipment Inventory table linespacing (fixes #186)

### LINKED ISSUE #186: Weapons & Equipment Inventory table content linespacing (in print)
In printed sheets produced by MegaMekLab, the weapon statistic lines below "Weapons & Equipment Inventory" (on the upper left of the sheet) are too close together, as it negatively impacts readability.

Significantly increasing line spacing would make the table much more legible, and it seems there is ample empty white space to allow for it.

pmjdebruijn:
In printed sheets produced by MegaMekLab, the weapon statistic lines below "Weapons & Equipment Inventory" (on the upper left of the sheet) are too close together, as it negatively impacts readability.

Significantly increasing line spacing would make the table much more legible, and it seems there is ample empty white space to allow for it.

### DIFF
--- a/src/megameklab/com/printing/PrintMech.java
+++ b/src/megameklab/com/printing/PrintMech.java
@@ -327,7 +327,7 @@ public class PrintMech extends PrintEntity {
         int currY = viewY + 10;
         
         float fontSize = FONT_SIZE_MEDIUM;
-        float lineHeight = getFontHeight(fontSize) * 0.8f;
+        float lineHeight = getFontHeight(fontSize) * 1.2f;
         
         addTextElement(canvas, qtyX, currY, "Qty", fontSize, "middle", "bold");
         addTextElement(canvas, nameX + indent, currY, "Type", fontSize, "start", "bold");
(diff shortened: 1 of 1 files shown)


pmjdebruijn:
Ok. Which Mech/Vehicle has this extremely long weapon list? I'd be happy to retune the distance so it'll fit just fine.

neoancient:
That was one of the variants of the Ares, which have some of the longest weapon lists I know of. But I didn't search for units with equipment lists at the various font size thresholds. But some math should be sufficient to ensure they fit. There's also the problem of the ammunition list not necessarily taking a set number of lines, so the bottom of the list is a moving target.

pmjdebruijn:
I've updated the pull request, with this code, I couldn't find any case where there was overlapping text.

However, these is an unhandled cornercase, that might already have been present.

Some weapons have a [Master] or [Clan] suffix, which caused the weapon type to wrap to the next line, and having multi-line weapon entries isn't great for readability either.

This seems to get "fixed" by FONT_SIZE_SMALL, for units with many weapon slots. However, Ideally there should be some logic where if a unit has long weapon types, the font size should be made smaller. As the readability impact of a slightly smaller font size seems to be less than having multi-line weapon slots.

But again, this is a pre-existing issue, my commit doesn't affect this one way or the other.

Issue #205, opened 2018-09-26
Blue Shield Particle Field Dampener (BSPFD) can't be dragged into ASF slots


UB-JM:
### Environment
Version 44.0, 
Windows 10 64 bit
Java 8

### Description
The Blue Shield Particle Field Dampener, an experimental IS system, can not be dragged and dropped into any slot of an aerospace fighter. Now, I don't know if it even should be, but it can be moved there through the RMB context menu.

PR #209, opened 2018-09-29
System wide equipment

neoancient:
Depends on MM system_wide_equipment. Adds a fuselage/hull location for aerospace units. Fighters automatically have ammo placed there when added. No weapon can be placed there, and non-fighters cannot place ammo there either.


neoancient:
This also has the side effect of fixing #205.

PR #216, opened 2018-10-27
One shot reprise

### LINKED ISSUE #213: Battle Armor Pop Up Mines way too heavy.
### Environment
0.45.1
Java 8
Win 10

### Description
The Battle Armor popup mine nominally weighs 200 kg. 
However, when equipping it, it takes roughly a 1000. Seems it's something to do with the ammo, as removing it leaves me with a slot of unallocated ammo of 4 shots, supposedly weighing -0.004 tons (yes, negative). Reducing the amount of shots reduces the weight on the squad by 200 kg each.

neoancient:
I think this is at least the third fix in as many months for issues surrounding one-shot launcher ammo. Previously the check for whether an ammo slot is one-shot is that is located in Entity.LOC_NONE and has a single shot. This ran into problems with ammo that has a single shot when full not showing up because it hasn't been allocated. This fix is prompted by BA popup mines, which fail the one-shot test because they have one shot for each trooper.

This should be the final fix, which avoids all the edge case exceptions by testing whether the ammo mount is linked by a one-shot launcher. 

Fixes #213

### DIFF
--- a/src/megameklab/com/ui/Aero/views/BuildView.java
+++ b/src/megameklab/com/ui/Aero/views/BuildView.java
@@ -128,10 +128,7 @@ public class BuildView extends IView implements ActionListener, MouseListener {
             }
         }
         for (Mounted mount : getAero().getAmmo()) {
-            if ((mount.getLocation() == Entity.LOC_NONE) && 
-                    ((mount.getUsableShotsLeft() > 1)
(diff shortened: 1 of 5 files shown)


neoancient:
I don't know why the diff is showing whole-file changes. I didn't make any whitespace changes, and my local repo is showing only the lines I changed as a diff, so I don't know what to do to fix it. I recommend using the option to hide the whitespace changes.

PR #243, opened 2019-01-07
Fix setting of manual BV.

neoancient:
Copy-paste error results in setting the entity intro year instead of the manual BV.

Also fixes the inability to clear the manual BV other than explicitly setting it to < 0, which is non-obvious.

### DIFF
--- a/src/megameklab/com/ui/util/IntRangeTextField.java
+++ b/src/megameklab/com/ui/util/IntRangeTextField.java
@@ -58,18 +58,32 @@ public class IntRangeTextField extends JFormattedTextField {
         setColumns(columns);
     }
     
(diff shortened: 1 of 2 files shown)


src/megameklab/com/ui/util/IntRangeTextField.java
```diff
@@ -87,7 +101,7 @@ public boolean verify(JComponent input) {
 
         @Override
         public boolean shouldYieldFocus(JComponent input) {
-            if (!super.shouldYieldFocus(input)) {
```
neoancient: The purpose of this change is to remove a deprecation warning in Java 9+.

Issue #238, opened 2018-12-22
Aero recordsheet not printing correctly


HammerGS:
### Environment
45.2 was reported, but having the same issue with the Repo
Windows 10
Java 8

### Description
Reported via PM.

Finally getting down to layout on the TRO...and I noticed one of my designs is missing dots in almost all areas. It is an Aero fighter with the following...

Front = 200 (this shows)
L/R Wings = 173 (these don't show)
Aft = 145 (these don't show)

Any thoughts? I'm using 0.45.2 now...and checked to see if it would display, and still doesn't. This file was created in a previous version (can't remember which) of MML.

Attached are the BLK files and a sample RS printed from the Repo.

### Files
[Habu.zip](https://github.com/MegaMek/megameklab/files/2705403/Habu.zip)
[Habu Prime.pdf](https://github.com/MegaMek/megameklab/files/2705404/Habu.Prime.pdf)

Dragon41673:
I'm guessing since no responses, no one has looked into this yet?

Thanks

neoancient:
While working on the wing and aft armor layout I noticed that the nose armor has a maximum of 200, which is exactly what this unit has. It's also blocky and in my opinion unsightly so I reworked that part also to use available space. This is what I have now:
![image](https://user-images.githubusercontent.com/16927464/51859927-ed035680-22fd-11e9-8dd3-461bcab94dd3.png)

Dragon41673:
That will work just fine...thank you!

neoancient:
Fixed by #246

PR #246, opened 2019-01-28
Asf armor pips

neoancient:
Fix for #238 
Band-aid rather than long term fix, but still an improvement. Armor layout is done by calculating all available pip positions and spreading the armor among them. When the amount of armor exceeds the number of available pip positions, the number of positions to advance between pips is calculated as zero and all the pips are drawn on top of each other. What I did is rework the layout code to use more space and simplify it. This doesn't remove the underlying problem that there is a maximum number of pips that can be drawn, but it increases the limit. The long-term solution is to move all units to the dynamic layout methods used when printing mech sheets, which can resize the pips as necessary to make them fit. But I haven't gotten to it yet.


Dragon41673:
Neo...thank you for getting this pushed through!

When will the updated MML be available to download that will have this fix?

Thank you!

PR #270, opened 2019-03-07
Fix #269 and improve support for loading relatively stored fluff images

### LINKED ISSUE #269: Fluff image path (imagefile) gets butchered when above MML base dir
### Environment
MegaMekLab 0.45.3
Windows 10
Java 1.8.0_201-b09

### Description
When setting a fluff image from a folder above the MegaMegLab directory, parts of the beginning of the path are missing in the saved mtf file.

lauxjpn:
This implementation can deal with relative paths above the MegaMekLab folder and fixes #269.
It also improves the support for loading fluff images that are stored relative to a MTF file they are referenced from.
This is an issue dear to my heart because I am using the lab mostly for my offline tabletop BattleTech/Rpg campaign, with different mech files and images for different sessions in their own folders.

The implementation from this pull request will first try to resolve the path of the image file in an absolute fashion.
Then it will try to load an image relative to the currently loaded MTF file (if it was indeed loaded explicitly from a file and not the cache).
Then it will try to load an image relative to the lab's "data/mechfiles/" path.
If that fails too, it will fall back to the previous algorithm and try to load an image from the fluffPath.

Saving wise, it will save the image path relative to the MTF file or the lab's "data/mechfiles/" path, whichever is shorter.

It should be backward compatible with previously saved files.

I did not find a path of the loaded file in the Entity class, so I added a CConfig property to keep track of the path of the currently loaded file. I would propose though, to keep the path (readonly) in the Entity itself to easier reference relatively stored files. The property could be empty for units loaded from the cache or another property could signal, if the Entity is a cached or not.


neoancient:
The only problem I see iine 247 of StatusBar, which incorrectly assumes that the selected file is within the MML distribution. Replacing that with a call to `Paths#relativize` should take care of it. The rest of the code changes proposed here create numerous new problems. If the path to the fluff image is not absolute, it is resolved relative to the working directory. Making it relative to the location of the unit file guarantees that it will not be found, unless the unit file happens to be in MML's root directory.

lauxjpn:
Thanks, I obviously missed the part with the working directory. So if I understand correctly, than I could run MML with a different working directory set where i could keep my fluff images.

For a more flexible image path resolving algorithm, wouldn't the following order to load an image file improve the current implementation, but still be backward compatible:
1. Absolute imagefile path
2. imagefile path relative to the current working directory
3. imagefile path relative to the current File
4. imagefile path relative to the current MML directory
5. Resolve by using unit name and model and searching the fluffpath

Ideally the shortest relative path and the absolute path to the image file would both be saved, so either of them could be resolved. With just one path being saved, I think the current choice to save a relative path is superior to an absolute approach.

The algorithm to choose a relative path to save with the file could basically be done in reverse of the loading path resolution algorithm:

1. If the user has explicitly chosen the image file from the fluffpath, and it would automatically resolve when loading by using unit name and model, then just don't save any imagefile path at all
2. In all other cases, choose the shortest of the following relative paths:
    a. imagefile path relative to the current working directory
    b. imagefile path relative to the current File
    c. imagefile path relative to the current MML directory

For the second case, saving the shortest relative path should ensure, that the path the user intended to put the fluff image relative to, is implicitly used.

Do you want me to split the original bugfix into it's own pull request, so it can be pulled right away and independent of the image path flexibility implementation?

neoancient:
I've been tied up for a couple weeks and haven't had a chance to get back with you until now.
Yes, submit the fix for relativizing paths correctly above the MML directory separately. The fix also needs to be able to handle the possibility that the two locations are on different drives on a Windows machine. I suspect attempting to relativize two paths on different drives will result in an IllegalArgumentException.

As for the rest of it, I don't understand what you're trying to do or why. The whole fluff image system assumes that the image will be relative to the MML directory. It's not very robust and was never intended to be. Anything more sophisticated should probably involve the possibility of embedding the image in the unit file to make it transferable to other computers.

neoancient:
Closing as outdated.

PR #332, opened 2019-08-08
EquipmentType constants

neoancient:
MML companion to MegaMek/megamek#1432
Plus some javadoc fixes.
Requires MM equipment_type_key branch.

### DIFF
--- a/src/megameklab/com/util/UnitUtil.java
+++ b/src/megameklab/com/util/UnitUtil.java
@@ -43,36 +43,7 @@ import javax.swing.ScrollPaneConstants;
 import javax.swing.text.DefaultCaret;
 import javax.swing.text.html.HTMLEditorKit;
 
-import megamek.common.Aero;
-import megamek.common.AmmoType;
-import megamek.common.BattleArmor;
-import megamek.common.Bay;
-import megamek.common.BipedMech;
-import megamek.common.CriticalSlot;
-import megamek.common.Dropship;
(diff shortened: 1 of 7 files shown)


src/megameklab/com/util/UnitUtil.java
```diff
@@ -845,7 +816,7 @@ public static Mounted getSingleCompactHeatSink(Mech unit) {
     }
 
     public static String getHeatSinkType(String type, boolean clan) {
-        String heatSinkType = "Heat Sink";
+        String heatSinkType;
```
neoancient: The initialized value is never used.

src/megameklab/com/util/UnitUtil.java
```diff
@@ -959,21 +930,22 @@ public static boolean isJumpJet(Mounted m) {
...
+     * @return     The {@link EquipmentType} lookup key for the jump jet
+     */
+    public static String getJumpJetType(int type) {
         if (type == Mech.JUMP_IMPROVED) {
-            if (clan) {
-                return "CLImprovedJump Jet";
-            } else {
-                return "ISImprovedJump Jet";
-            }
+            return EquipmentTypeLookup.IMPROVED_JUMP_JET;
```
neoancient: IS and Clan IJJs used to be different equipment because they have a different intro date, but they have been merged since the IO tech advancement update since they are functionally identical.

PR #340, opened 2019-08-22
Native dialogs

### LINKED ISSUE #174: Replace Java-based open dialog with OS native dialog
### Environment
Environment
MegaMekLab 0.44.0
OS X 10.13.5
Java 8 update 171

### Description
File -> Load -> From File uses a Java-based open dialog; it would be preferable to use OS Native dialogs.

neoancient:
Replaces use of Swing's cross-platform JFileChooser with the AWT FileDialog, which uses the OS-native dialog. This was already being used for saving and for opening fluff images.

This might best be reviewed one commit at a time. The first was a code cleanup. The next (after the upstream merge) involved tokenizing all the string literals. The third replaced the dialog.

I looked at replacing the cross-platform print dialog with a native one but that is a bit more involved and best left as a separate change.

Fixes #174

### DIFF
--- a/src/megameklab/com/util/MenuBarCreator.java
+++ b/src/megameklab/com/util/MenuBarCreator.java
@@ -1290,7 +1290,10 @@ public class MenuBarCreator extends JMenuBar implements ClipboardOwner {
     }
 
     private final FilenameFilter unitFilesFilter =
-            (dir, filename) -> filename.endsWith(".mtf")
-                    || filename.endsWith(".blk")
-                    || filename.endsWith(".hmp");
+            (dir, filename) -> {
+                String fn = filename.toLowerCase();
+                return fn.endsWith(".mtf")
+                        || fn.endsWith(".blk")
+                        || fn.endsWith(".hmp");
+            };
 }
--- a/src/megameklab/com/util/MenuBarCreator.java
+++ b/src/megameklab/com/util/MenuBarCreator.java
@@ -28,6 +28,7 @@ import java.awt.event.KeyEvent;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.PrintStream;
+import java.util.ResourceBundle;
 
 import javax.swing.BoxLayout;
 import javax.swing.JCheckBoxMenuItem;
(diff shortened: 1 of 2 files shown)


src/megameklab/com/util/MenuBarCreator.java
```diff
@@ -607,198 +591,119 @@ private void jMenuGetUnitWeightBreakdownFromCache_actionPerformed() {
...
-        JFileChooser f = new JFileChooser(filePathName);
-        f.setLocation(parentFrame.getLocation().x + 150, parentFrame.getLocation().y + 100);
-        f.setDialogTitle("Choose Unit");
-        f.setDialogType(JFileChooser.OPEN_DIALOG);
-        f.setMultiSelectionEnabled(false);
-
-        FileNameExtensionFilter filter = new FileNameExtensionFilter("Unit Files", "blk", "mtf", "hmp");
+    private @Nullable
+    File loadUnitFile() {
+        String filePathName = System.getProperty("user.dir") + "/data/mechfiles/";
```
sixlettervariables: Does Java have a path combiner? I know in .NET we'd do: `Path.Combine(System.getProperty("user.dir"), "data", "mechfiles")` and never care about path separators.

src/megameklab/com/util/MenuBarCreator.java
```diff
@@ -817,60 +722,58 @@ private void jMenuInsertImageFile_actionPerformed() {
 
             if (fDialog.getFile() != null) {
                 String relativeFilePath = new File(fDialog.getDirectory() + fDialog.getFile()).getAbsolutePath();
-                relativeFilePath = "." + File.separatorChar + relativeFilePath.substring(new File(System.getProperty("user.dir").toString()).getAbsolutePath().length() + 1);
+                relativeFilePath = "." + File.separatorChar + relativeFilePath.substring(new File(System.getProperty("user.dir")).getAbsolutePath().length() + 1);
```
sixlettervariables: Likewise, is there a relative path tool? (granted this code has probably worked just fine for years)

src/megameklab/com/util/MenuBarCreator.java
```diff
@@ -1390,4 +1289,8 @@ public void lostOwnership(Clipboard arg0, Transferable arg1) {
 
     }
 
+    private final FilenameFilter unitFilesFilter =
```
sixlettervariables: Do these endsWith need to be case sensitive?

neoancient: Good catch. I'll need to fix that.

src/megameklab/com/util/MenuBarCreator.java
neoancient: All I was doing here was removing the extraneous `toString()` call on a String expression. I didn't worry about the rest of the line, which has proved to be functional if not necessarily ideal.

src/megameklab/com/util/MenuBarCreator.java
neoancient: The java.nio.file.* package has tools that do that sort of thing, but I don't use it often enough to know off the top of my head. This line was moved from elsewhere in the code when I extracted a method to consolidate identical code that appeared 3-4 places in the class.

PR #353, opened 2019-10-06
Large engine crits

### LINKED ISSUE #352: Large engine crits not allocated correctly
### Environment
Ver 0.47.0

### Description
Well, it is a MegaMekLab issue, but since it is an equipment problem so I post it here as the other equipment problems does.

rr2

For normal Battlemechs, If you change the engine to the large engine, it does takes the latter two critical slots on the below of Center Torso, and that's fine, but if you put any equipments on there before raise its speed, MegaMekLab does not kicks the existing slots and there is no additional engine space in the Center Torso.

rr

For Superheavy Battlemechs, its additional slot on the engine is weirdly applied. As above, Superheavy also have a problem that does not kicks existing equipment slot. I think that it is better to put the last Engine slot just below of the Superheavy Gyro. Because Superheavy mechs are not able to choose the other types of Gyro, it seems not bad to set the fixed location when the mech is superheavy.

neoancient:
Fixes #352 
Large engines (rating > 400) occupy more critical slots and may need to punt equipment that has previously been installed there. The call to resetSystemCrits was missing in the method that recalculates engine rating. There was also a bug that assumed the large engine flag was set correctly in the engine combo box, but this was not always true. Also, superheavy mechs were reserving four slots for the gryo, when then should only be requiring two double slots.

### DIFF
--- a/src/megameklab/com/ui/view/MekChassisView.java
+++ b/src/megameklab/com/ui/view/MekChassisView.java
@@ -688,9 +688,12 @@ public class MekChassisView extends BuildView implements ActionListener, ChangeL
         if (null == e) {
             return null;
         }
-        // Clan and large flags are specific to the engine. the superheavy flag depends on the mech
-        // and may have changed since the last refresh.
-        int flags = e.getFlags() & (Engine.CLAN_ENGINE | Engine.LARGE_ENGINE);
+        // Clan and flag is specific to the engine. the superheavy and large flags depend on the mech
+        // and rating and may have changed since the last refresh.
+        int flags = e.getFlags() & Engine.CLAN_ENGINE;
+        if (getEngineRating() > 400) {
+            flags |= Engine.LARGE_ENGINE;
+        }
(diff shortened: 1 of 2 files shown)


src/megameklab/com/ui/view/MekChassisView.java
```diff
@@ -688,9 +688,12 @@ public Engine getEngine() {
         if (null == e) {
             return null;
         }
-        // Clan and large flags are specific to the engine. the superheavy flag depends on the mech
-        // and may have changed since the last refresh.
-        int flags = e.getFlags() & (Engine.CLAN_ENGINE | Engine.LARGE_ENGINE);
+        // Clan and flag is specific to the engine. the superheavy and large flags depend on the mech
```
sixlettervariables: "Clan **and** flag..." ?

neoancient: I can't read.

PR #358, opened 2019-10-22
MML Splash Screen

AaronGullickson:
As the title says, this PR adds an initial splash screen to MegaMekLab.

Screen Shot 2019-10-22 at 3 35 06 PM

The splash screen has the same basic look and feel as the MegaMek splash screen including the use of thematic buttons and fonts and resizing based on the user's screen resolution. In order to do this I had to add the the theme and font data to the MML data section. The buttons give the player the option of loading an existing unit (from the cache) or of starting a new unit design of a specific type. 

I also added a loading screen in between the pushing of the button and the full UI because I noticed there was a couple of seconds delay on my machine.

Screen Shot 2019-10-22 at 3 37 42 PM

Since I was using the loading screen here, I also changed all of the menu options in `MenuBarCreator.java` that load a unit of a different type to call up the loading dialog so the user will see a consistent loading screen regardless of how they load the unit. 

I also made a few adjustments to how the mainUI screen is displayed. It was displaying in the upper left corner and was typically too small, forcing scroll bars which was ugly and not user friendly IMO. So I created a finishSetup() routine in `MegaMekLabMainGUI.java` that does all of the work of centering on the screen and making sure that the main frame is large enough to fit everything without scrollbars when possible. 

Most of the work is done in `StartupGUI.java` and `LoadingDialog.java`.

~I am still having one issue that I am not sure is really an issue. When the user clicks the "Load a Unit" button, it will pull up the cache loading dialog the first time and then once units are loaded this will disappear and nothing else happens. If the user hits the button a second time, it works. This is not an issue when it is done through the menu bar rather than the splash screen, despite the code being identical. However, I don't get this behavior if I launch things with the debugger, so I am not convinced it is really an issue or just something with the cache being refreshed. I will make a note in the code where this is occurring.~ Fixed in commit 40f38aca.


AaronGullickson:
I should also note that the large number of files modified/added here is because of all the theme data being pulled in.

src/megameklab/com/MegaMekLab.java
```diff
@@ -26,160 +31,58 @@
...
-        showInfo();
-        
-        if (vehicle) {
-            Runtime runtime = Runtime.getRuntime();
-
-            getLogger().log(MegaMekLab.class, METHOD_NAME, LogLevel.INFO,
-                            "Memory Allocated [" +
-                            (runtime.maxMemory() / 1000) + "]");
-            // Need at least 200m to run MegaMekLab
-            if (runtime.maxMemory() < 200000000) {
```
AaronGullickson: I think this is very old code that was used to brute force past a memory leak issue. I doubt it is still an issue and if so this does not seem like a very good way to handle it.

src/megameklab/com/ui/Aero/MainUI.java
```diff
@@ -58,30 +54,13 @@
...
-    JPanel masterPanel = new JPanel();
-    JScrollPane scroll = new JScrollPane();
-    private MenuBarCreator menubarcreator;
 
     public MainUI(boolean primitive) {
 
         super();
         createNewUnit(Entity.ETYPE_AERO, primitive);
         setTitle(getEntity().getChassis() + " " + getEntity().getModel() + ".blk");
-        menubarcreator = new MenuBarCreator(this);
```
AaronGullickson: All of the stuff deleted from this point on in all of the MainUI methods is now being done in MegaMekLabMainGUI.finishSetup()

src/megameklab/com/ui/StartupGUI.java
```diff
@@ -0,0 +1,421 @@
...
+    }
+    
+    private void newUnit(long type, boolean primitive, boolean industrial, Entity en) {
+        frame.setVisible(false);
+        LoadingDialog ld = new LoadingDialog(frame, type, primitive, industrial, en);
+        ld.setVisible(true);
+    }
+    
+    private void loadUnit() {
+        UnitLoadingDialog unitLoadingDialog = new UnitLoadingDialog(frame);
```
AaronGullickson: This is the place where the first time the button is pushed it shows the UnitLoadingDialog and then disappears with nothing else happening.

AaronGullickson: Duh, check the log stupid. Here is where the problem seems to be:

```
Exception in thread "AWT-EventQueue-0" java.lang.NullPointerException
	at megamek.client.ui.swing.AdvancedSearchDialog$WeaponsTableModel.getColumnClass(AdvancedSearchDialog.java:1118)
	at javax.swing.table.TableRowSorter.useToString(TableRowSorter.java:241)
	at javax.swing.DefaultRowSorter.updateUseToString(DefaultRowSorter.java:624)
	at javax.swing.DefaultRowSorter.sort(DefaultRowSorter.java:573)
	at javax.swing.DefaultRowSorter.allRowsChanged(DefaultRowSorter.java:852)
	at javax.swing.JTable.notifySorter(JTable.java:4264)
	at javax.swing.JTable.sortedTableChanged(JTable.java:4124)
	at javax.swing.JTable.tableChanged(JTable.java:4401)
	at javax.swing.table.AbstractTableModel.fireTableChanged(AbstractTableModel.java:296)
	at javax.swing.table.AbstractTableModel.fireTableDataChanged(AbstractTableModel.java:198)
	at megamek.client.ui.swing.AdvancedSearchDialog$WeaponsTableModel.setData(AdvancedSearchDialog.java:1138)
	at megamek.client.ui.swing.AdvancedSearchDialog.populateWeaponsAndEquipmentChoices(AdvancedSearchDialog.java:1010)
	at megamek.client.ui.swing.AdvancedSearchDialog.<init>(AdvancedSearchDialog.java:378)
	at megamek.client.ui.swing.UnitSelectorDialog.<init>(UnitSelectorDialog.java:197)
	at megameklab.com.ui.StartupGUI.loadUnit(StartupGUI.java:379)
```

src/megameklab/com/ui/supportvehicle/SVMainUI.java
```diff
@@ -43,27 +42,13 @@
...
-        scroll.setViewportView(masterPanel);
-        scroll.setBorder(BorderFactory.createEmptyBorder());
-        this.add(scroll);
-
-        reloadTabs();
-        setVisible(true);
-        repaint();
-        refreshAll();
+        finishSetup();
         MechSummaryCache.getInstance();
```
AaronGullickson: @neoancient Some of the MainUI() methods had this line at the end, even though this is actually called at the very top of the constructor for MegaMekLabMainUI from which all of these inherit. Is it redundant or is something else going on here?

src/megameklab/com/ui/StartupGUI.java
AaronGullickson: Need to initialized EquipmentTypes. Fixed in commit 40f38aca.

NickAragua:
I gotta ask, what's the deal with the 755 files? Don't we usually handle that stuff by pulling it from the megamek repo at build-time?

AaronGullickson:
I am not sure how it works at build time, but if you want MML to run properly in eclipse then all those data files have to be there I think.

neoancient:
I don't have time to look at this tonight, but I think there should be some way to provide a reasonable default behavior if those files are missing.

AaronGullickson:
The easiest thing to do would be to just put the default skin data in MML as there is currently no way to change it from within MML anyway. Let me know what you want me to do.

AaronGullickson:
@neoancient @NickAragua I did some experimentation and I was able to remove most of the excess files for the skin themes, just leaving the defaults that are needed for the splash screen. That cut the number of file changes down to 33. 

This should allow the stand-alone MML build to work correctly. In the full MHQ build the user will have access to the full theme data and a change made in MM will then show up here as well through the mmconf.

AaronGullickson:
Ok, waiting for @neoancient feedback.

src/megameklab/com/ui/StartupGUI.java
```diff
@@ -0,0 +1,423 @@
...
+ */
+public class StartupGUI extends javax.swing.JPanel {
+
+    
+    private static final long serialVersionUID = 8376874926997734492L;
+    JFrame frame;
+    Image imgSplash;
+    BufferedImage backgroundIcon;
+    
+    private MegamekButton btnLoadUnit;
```
neoancient: The button fields aren't ever used outside initComponents() and can be converted to local variables.

src/megameklab/com/ui/MegaMekLabMainUI.java
```diff
@@ -112,10 +158,14 @@ public void exit() {
             CConfig.setParam("WINDOWSTATE", Integer.toString(getExtendedState()));
             // Only save position and size if not maximized or minimized.
             if (getExtendedState() == Frame.NORMAL) {
+                /*
```
neoancient: Why change this behavior? I consider being able to set the window size and location and have it come back in the same state the next time I run it to be desirable.

resources/megameklab/resources/Splash.properties
```diff
@@ -0,0 +1,15 @@
+btnLoadUnit.text=Load Existing Unit
+btnNewMek.text=New Mek
+btnNewVee.text=New Combat Vehicle
+btnNewSupportVee.text=New Support Vehicle
+btnNewProto.text=New Protomech
+btnNewBA.text=New Battle Armor
+btnNewPbi.text=New Infantry Unit
+btnNewAero.text=New Aero Fighter
+btnNewDropper.text=New Dropship/Small Craft
+btnNewLargeCraft.text=New Large Craft
```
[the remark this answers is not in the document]
AaronGullickson: I think capital ship is more natural than "advanced aerospace" units although its weird to  all a space station a capitol ship.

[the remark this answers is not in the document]
HammerGS: Might seem like an odd idea, but a button for Space Stations that jumps to the same starting point, but changes the drop to space station.  Or even button for JS,WS,SS that does that?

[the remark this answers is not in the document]
AaronGullickson: The problem is that we are pretty maxed out on buttons already and any space we have I would rather reserve for other future things that @neoancient will code for us 😉

src/megameklab/com/ui/MegaMekLabMainUI.java
AaronGullickson: The loading dialog now sets the window to be of large enough size to fit the whole tabbedpane without scrollbars (unless screen is too small) and centers it on the screen. I suppose we could leave the X and Y parameters here for placement, but my concern with saving the size components, is that the windows are different sizes for different unit types. So you might end up opening a window that is too small for the chosen unit.

src/megameklab/com/ui/supportvehicle/SVMainUI.java
neoancient: Actually having had time to thing about it I think the idea is to start the cache loading thread to cut or eliminate the wait the first time a unit is loaded from the cache. It would be better placed somewhere in the application startup process if it isn't already.

resources/megameklab/resources/Splash.properties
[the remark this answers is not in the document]
neoancient: I don't know if there's a really good solution for it. I raise it because being nitpicky can be asset when doing this sort of thing as long as you're not being a jerk about it (and I try not to be). If it goes out like this and people's feathers get ruffled it's no big deal to change it later. As far as what may get added in the future, I think we're down to mobile structures and handheld weapons, and HHWs may actually work better as part of mech construction.

src/megameklab/com/ui/MegaMekLabMainUI.java
neoancient: I work on a laptop with a max display resolution of 1366x768, so I don't have the luxury of being averse to scrollbars. I do sometimes size and position a window to take up part of the screen so I can see another window at the same time (in this case perhaps MML and a TRO). As long as I'm working with the same unit type it's not a problem, but if I switch unit types I think it's reasonable to expect MML not to override my choices and force me to go through the process of resizing the window repeatedly. I consider that a basic aspect of usability.

src/megameklab/com/ui/MegaMekLabMainUI.java
```diff
@@ -44,36 +50,76 @@
...
     public MegaMekLabMainUI() {
 
         EquipmentType.initializeTypes();
         MechSummaryCache.getInstance();
         new CConfig();
         UnitUtil.loadFonts();
-        System.out.println("Starting MegaMekLab version: " + MegaMekLab.VERSION);
-        
+        System.out.println("Starting MegaMekLab version: " + MegaMekLab.VERSION);        
         setLookAndFeel();
```
neoancient: This doesn't get called until after the main splash screen, which means that when you click the button to load the unit, the unit selector dialog can have a different look and feel than the the rest of the ui. I don't think it really belongs here in the first place, since it doesn't need to run every time the unit type changes, but now it needs to move to somewhere before the startup splash screen.

src/megameklab/com/ui/supportvehicle/SVMainUI.java
AaronGullickson: it is already done in the super() constructor.

src/megameklab/com/ui/MegaMekLabMainUI.java
neoancient: So after doing some work with the master branch I discovered that it actually doesn't do this now, and this whole line of argument is irrelevant. Never mind.

src/megameklab/com/ui/MegaMekLabMainUI.java
AaronGullickson: That should probably in the startup method of MegaMekLab.java. The System.out.println should also be there I think.

resources/megameklab/resources/Splash.properties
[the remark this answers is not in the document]
AaronGullickson: I changed it to "Capital-Scale Craft" which I think is accurate since these are all units that use capital scale armor.

src/megameklab/com/ui/StartupGUI.java
AaronGullickson: fixed in the latest

src/megameklab/com/ui/MegaMekLabMainUI.java
AaronGullickson: I moved `setLookandFeel()` and 'system.out.println` to MegaMekLab.startup()

PR #361, opened 2019-10-28
Move some initialization code earlier in the sequence.

neoancient:
The code moved only needs to run once, when the application is first launched. Having it in MegaMekLabMainUI caused it to run every time the unit type is changed. This doesn't hurt anything, but CConfig does need to initialize before calling setLookAndFeel().

### DIFF
--- a/src/megameklab/com/MegaMekLab.java
+++ b/src/megameklab/com/MegaMekLab.java
@@ -34,6 +34,8 @@ import javax.swing.UIManager;
 
 import megamek.MegaMek;
 import megamek.common.Configuration;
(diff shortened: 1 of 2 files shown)


src/megameklab/com/MegaMekLab.java
```diff
@@ -126,9 +129,12 @@ private static void showInfo() {
     }
     
     private static void startup() {
-        System.out.println("Starting MegaMekLab version: " + MegaMekLab.VERSION);
```
neoancient: This is already logged by showInfo(), which I moved earlier so it would run before initializing anything else.

PR #364, opened 2019-11-05
'Mech record sheet layout tweaks

neoancient:
There are some variations in the way text appears on record sheets which may be due to PDF print driver or differences in OS sizing the text differently. What works fine on my system is too large to fit in the available space when @HammerGS prints.

I've made the following changes to address this:
1. When positioning the "(CASE)" indicator, take into account that the location name is bold and takes up more space.
2. After calculating the size of font to use for the critical slots (as a proportion of the space available for the section), round it down. This seems to fix the differences in font sizes, which I suspect comes from a system call.
3. Fix a similar issue in the equipment table by making sure there is space between the Dmg and Min columns.
4. Trim leading and trailing spaces in the Dmg column entry. While working on the rest of this I noticed that a targeting computer shows as " [E]", with a leading space that causes it not to align with the rest of the column.

Example of something I printed:
Screenshot from 2019-11-05 12-37-33

Example of something printed using Windows and a different PDF driver:
Screenshot from 2019-11-05 12-39-07

### DIFF
--- a/src/megameklab/com/printing/PrintMech.java
+++ b/src/megameklab/com/printing/PrintMech.java
@@ -548,13 +548,14 @@ public class PrintMech extends PrintEntity {
         
         double rollX = viewX;
         double critX = viewX + viewWidth * 0.11;
+        double critWidth = viewX + viewWidth - critX;
         double gap = 0;
         if (mech.getNumberOfCriticals(loc) > 6) {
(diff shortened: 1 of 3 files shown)


neoancient:
StringUtils has whitespace issues. View with whitespace changes hidden to see the one minor change.

PR #378, opened 2019-11-30
Color record sheets

neoancient:
This is the first installment (the Mech sheets) of a project to rework the record sheets in accordance with the following goals:

1. Make the size and placement of the various elements of the sheet consistent across the various types. Currently the margins and the size and placement of the logo and various panels has enough variance that when you page through a RS book you can see them jumping around.
2. Make future changes easier. These templates are machine generated, and it's much easier to alter the generation code and update the templates than to edit them individually.
2. Eliminate cruft from the record sheet SVG templates. These are exports from the blank record sheets in the rule book pdf. The new ones include only the relevant data.
3. Add color. The BT and CGL logos and the heat scale have been replaced by color versions.
4. Replace all the record sheet printing code with the MVC approach currently used by mechs, conventional infantry, and advanced aero.

Some code changes were needed to reflect changes in the template model. There is also a fix to a current bug that SI pips are not shown on the LAM sheet, and some code cleanup.

Sample output:
[Mad_Cat__Timber_Wolf__Prime_etc__inux-generated_files-job_1134.pdf](https://github.com/MegaMek/megameklab/files/3907540/Mad_Cat__Timber_Wolf__Prime_etc__inux-generated_files-job_1134.pdf)

The template generation code is here:
https://github.com/neoancient/rstemplate


src/megameklab/com/printing/PrintMech.java
```diff
@@ -366,7 +386,7 @@ protected void drawArmorStructurePips() {
                 element = getSVGDocument().getElementById("isPips" + mech.getLocationAbbr(loc));
                 if (null != element) {
                     addPips(element, mech.getOInternal(loc),
-                            (loc == Mech.LOC_HEAD) || (loc == Mech.LOC_CT) || (loc == Mech.LOC_CLEG));
```
neoancient: The diff shown here leaves off the relevant condition that skips this block when loc > Mech.LOC_HEAD.

src/megameklab/com/printing/PrintEntity.java
```diff
@@ -127,7 +128,7 @@ protected void printImage(Graphics2D g2d, PageFormat pageFormat, int pageNum) {
         drawArmor();
         drawStructure();
         Element eqRect = getSVGDocument().getElementById("inventory");
-        if ((null != eqRect) && (eqRect instanceof SVGRectElement)) {
+        if (eqRect instanceof SVGRectElement) {
```
NickAragua: Is it possible for eqRect to be null?

neoancient: Yes, but in that case `eqRect instanceof SVGRectElement` will be false. So the null check is redundant.

neoancient: And by "possible" I mean "only if someone screws around with the template." And I'm including myself in that "someone."

NickAragua: Makes sense.

PR #394, opened 2020-01-13
Fix primitive jumpship armor rounding per errata.

### LINKED ISSUE #386: Armor Calculation Issue with Primitive Jumpship (Rounding issue?) (Re-Opened for comment)
From the Discord Conversation

> I did the calculations for one of Liam's Ghosts designs by hand earlier so I suppose I can use that
> Its a 300,000 ton ship with 50 SI. 236 tons of primitive armor.
>  
> Total armor value of: 82.
>  
> My calculations say that 236 tons on such would be 81 armor.
>  
> 236 x 0.4 = 94.4, rounded down to 94.
> 94 + 30 [SI provided armor] = 124
> 124x0.66 = 81.84, rounded down to 81.
> [as the example in Interstellar Operations does it]
> As far as I can tell, MML is doing 94.4+30x0.66=82.1
> 
> I think the issue is not many people actually have experience doing large craft by hand so are able to pick these out. The other potential reason MML isn't doing primitive armor properly is that it might be multiplying whatever SO armor value it has by the primitive modifier

**From the 1.1 Errata Doc**
Primitive JumpShips use the rules and limits for standard WarShips when computing their armor values (see p. 152, SO), and may not mount any of the advanced armor types such as improved ferro-aluminum, ferro-carbide, or lamellor ferro-carbide. This means that the maximum tonnage of armor a Primitive JumpShip may carry is equal to the ship’s SI tonnage, divided by 50 (rounded down to the nearest half ton). In addition to this, as with modern WarShips, Primitive JumpShips will receive additional “weight-free” armor points per facing, based on their structural integrity values. All
armor mounted on a Primitive JumpShip—including the “free” armor provided by the vessel’s structural integrity value— must be multiplied by 0.66, rounding down to the nearest whole number.

Primitive JumpShip Construction [example text] (p. 131)
1) Under “Step 4: Add Armour”, replace the example paragraph with the following:
Reviewing the Advanced Aerospace Unit Armor Table lets David know that he could assign a maximum of 20 tons of standard armor to his ship [1,000 (SI mass) ÷ 50 = 20, no rounding required]. Even though this is a low amount, he decides he’s going to assign only 19 tons of Primitive armor.
Before placing the armor, he first determines how much “weight-free” armor the Aquilla will receive in addition to this, which is 1 point of capital-scale armor per facing, for a total of 6 points [10 (SI) ÷ 10 = 1 per facing; 6 facings x 1 point per facing = 6 points].

David then multiplies the 19 tons of armor he has opted for by the standard Advanced Aerospace Unit Armor Weights for a 100,000-ton Inner Sphere WarShip using standard armor, and finds that this yields 15 armor points [19 (armor tonnage) x 0.8 (standard armor for a 100,000 vessel) = 15.2, rounded down to 15]. He then adds the free 6 points to 16 for a total of 21. Multiplying this total armor value by the Primitive armor factor of 0.66, he finds he has 13 armor points to allocate [21 (non-Primitive armor points) x .66 (Primitive armor factor) = 13.86, rounding down to 13].

He thus assigns this armor as follows: 3 to the nose, 2 to each fore-side, 2 to each aft-side and 2 to the aft.
This leaves a running total of 66,859 tons.
### Environment
What version of MekLab does your issue apply to?  Current to repo (47.3)
What operating system are you using?  Windows

### Description
From MML 47.3 - Remade the Aquilla to follow the Example in the IO + Errata
image

neoancient:
As clarified in the errata to IO, the armor points for a primitive jumpship are calculated as for standard armor, then rounded down, then the primitive multiplier of 0.66 is applied, then rounded down again. The free armor points gained from SI are rounded normally as for standard advanced aero, but the 0.66 mulitplier is applied to that value as well and rounded down. The way MML currently does it is by applying the primitive factor to the standard points per ton then rounding once, but this can carry over enough accumulation of fractional amounts to result in an extra point or two. The double rounding required special handling.

Neither IO nor the errata address the method used for dropships, but since IS standard armor is always an even whole number per ton, each half ton lot is a whole number of armor points and the first rounding step is not required.

Fixes #386

### DIFF
--- a/src/megameklab/com/util/UnitUtil.java
+++ b/src/megameklab/com/util/UnitUtil.java
@@ -1500,6 +1500,12 @@ public class UnitUtil {
                     EquipmentType.getProtomechArmorWeightPerPoint(unit.getArmorType(Protomech.LOC_TORSO)));
         } else if (unit.isSupportVehicle()) {
             return Math.floor(armorTons / TestSupportVehicle.armorWeightPerPoint(unit));
+        } else if ((unit instanceof Jumpship)
+                && unit.getArmorType(unit.firstArmorIndex()) == EquipmentType.T_ARMOR_PRIMITIVE_AERO) {
+            // Because primitive jumpship armor has an extra step of rounding we have to give it special treatment.
+            // Standard armor value is computed first, rounded down, then the primitive armor mod is applied.
+            return Math.floor(Math.floor(armorTons * TestAdvancedAerospace.armorPointsPerTon((Jumpship) unit,
+                    EquipmentType.T_ARMOR_AEROSPACE, false)) * 0.66);
         }
         return armorTons * UnitUtil.getArmorPointsPerTon(unit,
                 unit.getArmorType(1), unit.getArmorTechLevel(1));
@@ -1520,7 +1526,7 @@ public class UnitUtil {
         } else if (entity.hasETypeFlag(Entity.ETYPE_JUMPSHIP)) {
             double points = Math.round(((Jumpship) entity).getSI() / 10.0) * 6;
             if (((Jumpship) entity).isPrimitive()) {
-                return points * EquipmentType.getArmorPointMultiplier(EquipmentType.T_ARMOR_PRIMITIVE_AERO);
+                return Math.floor(points * EquipmentType.getArmorPointMultiplier(EquipmentType.T_ARMOR_PRIMITIVE_AERO));
             } else {
                 return points;
             }


HammerGS:
There might still be an issue with this.  Was sent this on Discord:

```
Test Design:
300,000 Tons, 20 Structural Integrity, Primitive Armor max at 120 Tons
Provides 38 Armor Points. Says at max armor it should have 39 Armor Points.
 
My own math for design, using method in Interstellar Operations
120 Tons multiplied by value for tonnage = 48 Armor
20 SI provides 12 Armor.
60 Armor multiplied by primitive value [of 0.66] = 39 Armor.
 
System used in MML seems to multiply the armor tonnage directly by 0.66 and the SI provided armor by 0.66, then combine them together. I am not sure if its rounding down at any point there.
MML says 0.66 * 0.4 = 0.26, but it is 0.264, if armor tonnage value was 0.264, assuming no rounding at any point in procedure prior to final value it would generate 39.

If anyone wants to follow up with the person the link to the discord is:
https://discord.gg/D9jFn52 in the Megamek channel, and look for post with todays date from VhenRa.

```

PR #398, opened 2020-01-15
Vehicle Record Sheet Overhaul

### LINKED ISSUE #42: Can't print naval vehicles
Missing armor pip SVG images for naval vehicles.

neoancient:
Implements the SVG template-based printing method currently used for 'Mechs, conventional infantry, and JS/WS/SS. Some things worth particular note:
1. I wrote an all-new pip layout algorithm that is less rigid than the one I wrote a few years ago for superheavy 'Mechs. The results come much closer to the canon layout patterns used by biped 'Mechs. Coding for aesthetics is much more involved than making it just work, so ArmorPipLayout.java is rather involved but hopefully not too obscure.
2. I've replaced the String literals used as SVG element ids with a set of named constants.
3. I reworked the way PrintRecordSheet initializes the SVG document to make it more flexible. This allows the non-naval vehicle sheets, which have either two units or a unit and a set of tables, to create a base document in memory and insert the two appropriate templates.
4. While creating superheavy naval vehicles to test, I discovered a bug in the way new vehicle entities are initialized and fixed it. This change makes this branch dependent on Megamek/megamek#1605

Here is a sample of the results. There are a few at the end that are not fully constructed units but show what the sheets look like for naval/WiGE dual turrets and superheavy naval combat vehicles, none of which are represented in the units in MM:

[Skulker_Wheeled_Scout_Tank__Standard__etc__ated_files-job_1193.pdf](https://github.com/MegaMek/megameklab/files/4067299/Skulker_Wheeled_Scout_Tank__Standard__etc__ated_files-job_1193.pdf)

Fixes #42


data/images/recordsheets/naval_dualturret_standard.svg
```diff
@@ -0,0 +1,1495 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
sixlettervariables: I normally wouldn't review these by hand, but the files are formatted quite bizarrely.

src/megameklab/com/printing/PrintTank.java
```diff
@@ -0,0 +1,212 @@
...
+        sb.append(" Record Sheet");
+        return sb.toString();
+    }
+
+    @Override
+    protected void writeTextFields() {
+        super.writeTextFields();
+        setTextField(MOVEMENT_TYPE, tank.getMovementModeAsString());
+        setTextField(ENGINE_TYPE, tank.getEngine().getEngineName()
+                .replaceAll("\\d+|\\[.*]", "").trim());
```
sixlettervariables: Reading this Regex, is this intended to match any one or more numbers or the literal `[.*]`?

data/images/recordsheets/naval_dualturret_standard.svg
neoancient: Yes, they're built in memory using DOM then written to a file using Batik's SVGGraphics2D#stream. I haven't found a way to influence the formatting of the output, but I haven't spent a lot of time looking.

src/megameklab/com/printing/PrintTank.java
neoancient: One or more numbers or anything in literal brackets. So "210 XL [Tank Engine]" becomes " XL " which then has the leading and trailing space trimmed.

sixlettervariables: Sounds good.

PR #401, opened 2020-01-21
FighterChassis and Aero StructureTab refactoring

Windchild292:
This is a minor refactor that I did while looking at the code relating to VSTOL. The .gitignore change is needed for those of us who use the Idea IDE

### DIFF
--- a/src/megameklab/com/ui/Aero/tabs/StructureTab.java
+++ b/src/megameklab/com/ui/Aero/tabs/StructureTab.java
@@ -181,11 +181,9 @@ public class StructureTab extends ITab implements AeroBuildListener, ArmorAlloca
 
         panSummary.refresh();
         addAllListeners();
-
     }
 
     public JLabel createLabel(String text, Dimension maxSize) {
-
         JLabel label = new JLabel(text, SwingConstants.RIGHT);
 
         setFieldSize(label, maxSize);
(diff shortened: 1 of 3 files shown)


src/megameklab/com/ui/Aero/tabs/StructureTab.java
```diff
@@ -443,15 +437,12 @@ public void maximizeArmor() {
...
+        double currentTonnage = UnitUtil.getEntityVerifier(getAero()).calculateWeight();
         currentTonnage += UnitUtil.getUnallocatedAmmoTonnage(getAero());
         double totalTonnage = getAero().getWeight();
-        double remainingTonnage = TestEntity.floor(
-                totalTonnage - currentTonnage, TestEntity.Ceil.HALFTON);
+        double remainingTonnage = TestEntity.floor(totalTonnage - currentTonnage, TestEntity.Ceil.HALFTON);
         
-        double maxArmor = Math.min(getAero().getArmorWeight() + remainingTonnage,
-                UnitUtil.getMaximumArmorTonnage(getAero()));
+        double maxArmor = Math.min(getAero().getArmorWeight() + remainingTonnage, UnitUtil.getMaximumArmorTonnage(getAero()));
```
neoancient: This line is too long. Per the Google Java Style Guide, which our style guide is based on, columns should be limited to 100 characters. I wouldn't apply it so strictly that I would break up strings, but longer lines require more eye movement and are harder for people like me who work on older displays that can't fit the whole line on the screen at once.

Windchild292: I set up a line that is at 100 characters on my IDE, so I can view it. I just missed it in this case. Reverted, and my apologies.

Found out why, my laptop is set to 100 while my desktop is at 120. Fixed the line on my desktop so this doesn't happen again.

PR #409, opened 2020-01-28
Print inventory refactor

### LINKED ISSUE #315: Hull equipment on WS Record Sheet
### Environment
megameklab-windows-0.45.4
Java version 1.8.0_211

### Description
Hull equipment (Naval-Com Scanner Suite, HPG, etc), Docking Collars and Lithium-Fusion Batteries aren't shown on the WarShip Record Sheet. I suggest adding a notes section or hull equipment section after the section for Standard Scale stuff (see page two in example file)

[Alexandria CoreShip.pdf](https://github.com/MegaMek/megameklab/files/3293633/Alexandria.CoreShip.pdf)

neoancient:
I started this in the process of reworking aerospace record sheets with the intention of reducing redundant code used to display the weapons and equipment inventory. Though I'm currently working on fighters and small craft, I wanted to make sure the system would be able to handle large craft weapon bays as well, so I ended up significantly reworking the inventory printing code and moved it into the same utility.

Currently the inventory for mechs and vehicles is handled by PrintEntity#writeEquipment, which is a long method that needs to be broken down anyway. I moved it into a utility class called InventoryWriter, where it could be refactored into smaller methods. The text for the various fields comes from the EquipmentInfo and RecordSheetEquipmentLine, which perform functions that are not entirely distinct, so I consolidated them into the StandardInventoryEntry class. It shares a common interface with WeaponBayInventoryEntry, which is code factored out of PrintCapitalShip. The older classes can be removed once the aerospace sheet printing overhaul is complete and they are no longer needed.

Along the way I corrected several oversights and inconsistencies:
1. Large craft weapon bays will now show AV adjusted for Artemis IV/V and PPC capacitors (which also affect the bay heat value).
2. The presence of Artemis/Apollo in large craft bays is now indicated on the sheet as a footnote to the standard scale block, including the to-hit modifiers for Artemis V and Apollo.
3. PPC capacitors are listed with the weapon they modify rather than as a separate piece of equipment.
4. Large craft now show non-bay equipment in the hull location (fixes #315).
5. Account for variable AV in MML and ATM bays due to munition alternatives.


src/megameklab/com/printing/InventoryEntry.java
```diff
@@ -0,0 +1,156 @@
...
+            } else if (weapon instanceof SRMWeapon) {
+                return 2;
+            }
+        } else if (linkedBy.hasFlag(MiscType.F_ARTEMIS_PROTO) && weapon.getRackSize() == 2) {
+            // The +1 only makes a difference for SRM2
+            return 2;
+        } else if (linkedBy.hasFlag(MiscType.F_APOLLO) && weapon.getRackSize() == 2) {
+            // The -1 only makes a difference for SRM2
+            return -2;
+        } else if (bay && linkedBy.hasFlag(MiscType.F_PPC_CAPACITOR)) {
```
sixlettervariables: The comments mention +1 and -1, but they return +2 and -2. Is this intended? (I'm unfamiliar with this area of the code)

neoancient: Yeah, that's confusing. The +/- 1 is to the cluster hits table. The +/- 2 is to the damage because of one more missile hit on average. But that doesn't seem right because that would be a difference of two missIles between 6 and 8. I'll double check that when I get home.

neoancient: I think I was looking at the wrong pair of rows. The -1 cluster hit mod for Apollo doesn't affect SRM2, but it does reduce the AV for SRM4. I fixed that and made the comments actually helpful.

src/megameklab/com/printing/InventoryWriter.java
```diff
@@ -486,9 +486,9 @@ private int calcLineCount(float fontSize, double viewWidth) {
             }
             lines += rows;
         }
-        lines += sheet.getTextLength(ammoText, fontSize) / viewWidth;
-        lines += sheet.getTextLength(featuresText, fontSize) / viewWidth;
-        lines += sheet.getTextLength(quirksText, fontSize) / viewWidth;
+        lines += Math.ceil(sheet.getTextLength(ammoText, fontSize) / viewWidth);
+        lines += Math.ceil(sheet.getTextLength(featuresText, fontSize) / viewWidth);
+        lines += Math.ceil(sheet.getTextLength(quirksText, fontSize) / viewWidth);
```
neoancient: While testing the aero record sheet branch I discovered that the font scaling was only counting full lines.

src/megameklab/com/printing/InventoryEntry.java
neoancient: So apparently I was doing this section on autopilot and didn't actually think about it until now. Apollo is for MRMs, which comes in rack sizes of 10, 20, 30, and 40. It has no affect on the aerospace AV. I fixed it properly this time.

PR #413, opened 2020-02-01
Aerospace record sheet facelift, part 1

neoancient:
Spiffification of record sheets for aerospace and conventional fighters, fixed wing support vehicles (which use conventional fighter sheets), and small craft. Dropships and other large craft are deferred to a separate update.

A preview of some of the changes:
1. The aerospace armor diagrams have more background detail than mechs and vehicles, so I increased the width of the pip stroke and added a white fill.
2. I moved the heat sink pip drawing method from PrintMech to PrintEntity where it can be shared with aerospace fighters and small craft.
3. I moved the transport bay formatting code from PrintCapitalShip to InventoryWriter, where it can be shared by small craft.
4. I added a new utility method to PrintRecordSheet that creates a rounded rectangle, which is used for the external stores section.
5. I moved support vehicle basic and advanced fire control from the equipment table to the features section, which is in keeping with how it is shown in TROs.
6. I fixed a bug in the armor and structure pip layout code that can cause an endless loop when with an odd number of pips in a section that is divided in half horizontally, such as the conventional fighter nose section (which flanks the cockpit).

Here's the sample:
[aerospace sample.pdf](https://github.com/MegaMek/megameklab/files/4142809/aerospace.sample.pdf)


src/megameklab/com/printing/InventoryWriter.java
```diff
@@ -144,6 +146,8 @@ public InventoryWriter(PrintEntity sheet) {
         this.standardBays = new ArrayList<>();
         this.bayFootnotes = new ArrayList<>();
         this.ammo = new TreeMap<>();
+        this.transportBays = sheet.getEntity().getTransports().stream().filter(t -> t instanceof Bay)
```
NickAragua: A little difficult to decode. "All transports that are bays and are not living quarters"?

NickAragua: Also, is there any point at which this variable becomes stale within this context? I imagine not, since it looks like it's a print action, but...

src/megameklab/com/printing/PrintAero.java
```diff
@@ -0,0 +1,269 @@
+/*
+ * MekBuilder - unit design companion of MegaMek
```
sixlettervariables: Old header?

src/megameklab/com/printing/PrintAero.java
```diff
@@ -0,0 +1,269 @@
...
+        if (aero.hasWorkingMisc(MiscType.F_ADVANCED_FIRECONTROL)) {
+            sj.add("Advanced Fire Control");
+        } else if (aero.hasWorkingMisc(MiscType.F_BASIC_FIRECONTROL)) {
+            sj.add("Basic Fire Control");
+        }
+        Map<String, Double> transport = new HashMap<>();
+        for (Transporter t : aero.getTransports()) {
+            if (t instanceof TroopSpace) {
+                transport.merge("Infantry Bay", t.getUnused(), Double::sum);
+            // include carge bays for fighters and fixed wing, but small craft get a block for transport bays
```
sixlettervariables:
```suggestion
            // include cargo bays for fighters and fixed wing, but small craft get a block for transport bays
```

src/megameklab/com/printing/PrintMech.java
```diff
@@ -533,51 +533,6 @@ protected void drawFluffImage() {
         }
     }
 
-    private void drawHeatSinkPips(SVGRectElement svgRect) {
```
NickAragua: How's the heat sink pips look now for mechs?

src/megameklab/com/printing/InventoryWriter.java
neoancient: 1. Yes
2. No. This instance only exists for a single print action.

src/megameklab/com/printing/PrintAero.java
neoancient: Experimental side project. I thought I was changing the new file template for the project, but it turned out it was global. I've changed it back now, but I missed this one.

src/megameklab/com/printing/PrintMech.java
neoancient: It's in a series of columns in the heat block. This doesn't change the appearance. It just moves it into the parent class so it can be available to both PrintMech and PrintAero.

PR #423, opened 2020-02-07
Aerospace record sheet facelift, part 2

neoancient:
This is the second half of the aerospace record sheet overhaul. This incorporates all the changes in #413, so would be easier to digest after that one is merged.

Some notes about changes:
1. Breaks up the older PrintCapitalShip, which previously was a subclass of PrintEntity, into PrintDropship, which subclasses PrintAero, and PrintCapitalShip, which subclasses PrintDropship. PrintDropship has the code that is common to all large craft, particulary involving sorting out the various blocks of text in the inventory panel and making sure the font is scaled appropriately to fit everything. PrintCapitalShip is responsible for drawing the armor and internals blocks, which are done in 10x10 grids rather than circular pips.
2. Removed some code that is not used, including merging identical fore and aft weapon bays into a single line, and allowing for the possibility of vertical alignment of JS/WS/SS internal pip blocks.
3. Moved non-weapon equipment from a table that is designed to display weapons to a block of text at the bottom of the inventory panel.
4. Fixed a problem that causes jumpship aft armor on top of the ship outline.

Samples:
[Leopard__3056__etc-job_1283.pdf](https://github.com/MegaMek/megameklab/files/4169061/Leopard__3056__etc-job_1283.pdf)


neoancient:
Closing due to GitHub bug showing an excessive diff.

PR #434, opened 2020-02-09
Expand theme support and include dark themes

sixlettervariables:
This expands the theme support in MML to support dark themes and reworks most of the underlying controls to be look-less to ensure they display correctly under each theme. This isn't 100% perfect, but pretty close for an hour of work.
image
image
image
image

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -75,9 +75,6 @@ dependencies {
     // the conversion to batik is complete.
     implementation files('lib/svgSalamander.jar')
 
-    // Light and Dark themes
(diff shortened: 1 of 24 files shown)


build.gradle
```diff
@@ -75,6 +75,9 @@ dependencies {
     // the conversion to batik is complete.
     implementation files('lib/svgSalamander.jar')
 
+    // Light and Dark themes
+    implementation 'com.formdev:flatlaf:0.26'
```
neoancient: Dependencies are transitive, so this is redundant because it's inherited from the MM dependency.

PR #439, opened 2020-02-11
437: Fixing Space Station Display Transport Bay Construction Issues

### LINKED ISSUE #437: Disappearing DS Collars on Space Station
Sent via PM

Think I found a bug, but cannot find anything in the logs. Using 47.3 MekHQ, MekLab

 _Take Olympus Recharge Station
 Remove cargo bay
 Modify all pressure bay to 1M ton (1 NOS, 1 ALS, 1 ARS)
 Add 3 Recharge Batteries to hull
 Should still have 4 dropship collars showing
 Use remaining cargo
 Dropship collars zero out and cannot be selected_

 _I have duplicated this repeatedly_

Can can duplicate the behaviour in 47.4
### Environment
What version of MekLab does your issue apply to?  Standalone MML
What operating system are you using?  Windows

Log files not available

Windchild292:
This fixes #437.


neoancient:
This fixes part of the problem, but it's still possible to set the repair facility capacity above the maximum.

Windchild292:
Based on my reading of the rules, they only reduce the number allowed. They do not require them to be built.

neoancient:
I can see that as a valid reading.

PR #485, opened 2020-02-21
Ignore OS launchers when filtering ammo on equipment tab.

### LINKED ISSUE #479: Additional tons of ammunition for one-shot missile launchers available
Using MML 0.47.4 on OS X, once I've added a one-shot missile launcher to a unit, I have the option to add additional ammunition for the corresponding regular launcher. So, adding an LRM 5 (OS) makes LRM 5 ammo available to be added.

neoancient:
There's even an argument to ignore oneshot. It just wasn't being used for most unit types.
Fixes #479


Windchild292:
Given how all the uses of canUseAmmo(Entity unit, AmmoType atype) are being removed, do we need to keep that method around?

neoancient:
No. Removed.

PR #491, opened 2020-02-23
Large craft rs

neoancient:
This is the second half of the aerospace record sheet overhaul. Some notes about changes:

Breaks up the older PrintCapitalShip, which previously was a subclass of PrintEntity, into PrintDropship, which subclasses PrintAero, and PrintCapitalShip, which subclasses PrintDropship. PrintDropship has the code that is common to all large craft, particulary involving sorting out the various blocks of text in the inventory panel and making sure the font is scaled appropriately to fit everything. PrintCapitalShip is responsible for drawing the armor and internals blocks, which are done in 10x10 grids rather than circular pips.

Removed some code that is not used, including merging identical fore and aft weapon bays into a single line, and allowing for the possibility of vertical alignment of JS/WS/SS internal pip blocks.

Moved non-weapon equipment from a table that is designed to display weapons to a block of text at the bottom of the inventory panel.

Fixed a problem that causes jumpship aft armor on top of the ship outline.

Samples:
[Leopard__3056__etc-job_1283.pdf](https://github.com/MegaMek/megameklab/files/4240547/Leopard__3056__etc-job_1283.pdf)


src/megameklab/com/printing/PrintRecordSheet.java
```diff
@@ -268,14 +269,15 @@ public int print(Graphics graphics, PageFormat pageFormat, int pageIndex) throws
             node.paint(g2d);
             /* Testing code that outputs the generated svg
             try {
-                Transformer transformer = TransformerFactory.newInstance().newTransformer();
-                Result output = new StreamResult(new File("out.svg"));
-                Source input = new DOMSource(svgDocument);
+                javax.xml.transform.Transformer transformer = javax.xml.transform.TransformerFactory.newInstance().newTransformer();
+                javax.xml.transform.Result output = new javax.xml.transform.stream.StreamResult(new File("out.svg"));
+                javax.xml.transform.Source input = new javax.xml.transform.dom.DOMSource(svgDocument);
```
neoancient: I changed this so I wouldn't have to keep messing with the imports when I enable this to troubleshoot the svg output.

PR #511, opened 2020-02-28
No dual turrets for WiGE vehicles.

### LINKED ISSUE #505: WiGE units should not be able to mount dual turrets
Using MML 0.47.4 under OS X with Java build 1.8.0_51-b16, I can put dual turrets on a WiGE. TO p. 347 says dual turrets can't be put on VTOL, WiGE, fixed-wing, airship, or satellite units. VTOLs and fixed-wing units work just fine, and neither satellites nor airships are implemented yet.

neoancient:
Per TacOps, p. 347 (2018 printing), WiGE vehicles cannot mount dual turrets. This is a change from previous printing.

Fixes #505

### DIFF
--- a/src/megameklab/com/ui/view/CVChassisView.java
+++ b/src/megameklab/com/ui/view/CVChassisView.java
@@ -330,7 +330,7 @@ public class CVChassisView extends BuildView implements ActionListener, ChangeLi
             }
         } else {
             cbTurrets.addItem(TURRET_SINGLE);
-            if (techManager.isLegal(TA_DUAL_TURRET)) {
+            if (!EntityMovementMode.WIGE.equals(getMovementMode()) && techManager.isLegal(TA_DUAL_TURRET)) {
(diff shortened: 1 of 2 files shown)


src/megameklab/com/ui/view/CVChassisView.java
```diff
@@ -330,7 +330,7 @@ private void refreshTurrets() {
             }
         } else {
             cbTurrets.addItem(TURRET_SINGLE);
-            if (techManager.isLegal(TA_DUAL_TURRET)) {
+            if (!EntityMovementMode.WIGE.equals(getMovementMode()) && techManager.isLegal(TA_DUAL_TURRET)) {
```
NickAragua: I'll assume that the techManager.isLegal() method doesn't have access to the movement mode.

neoancient: The tech manager only deals with intro years, tech base, and tech level.

PR #544, opened 2020-03-28
Infantry Record Sheet rework

neoancient:
Moves conventional infantry record sheet printing to the new framework, using Batik as the SVG library instead of svgSalamander. We are currently distributing with a custom build of svgSalamander that has some fixes to text layout, but with the infantry record sheets no longer using it, the release is good enough for loading the SVG images still used by ProtoMechs and BattleArmor.

The infantry sheets use two new classes in the printing package. PrintSmallUnitSheet is responsible for the overall page, including the header and footer. It uses PrintInfantry for each platoon or point up to four per page. The same class will also handle individual BA points/squads and protomechs as those sheets get redesigned.

Sample:
[Foot_Platoon__MG__etc__sing_non-Linux-generated_files-job_1315.pdf](https://github.com/MegaMek/megameklab/files/4397430/Foot_Platoon__MG__etc__sing_non-Linux-generated_files-job_1315.pdf)


src/megameklab/com/printing/PrintInfantry.java
```diff
@@ -0,0 +1,440 @@
...
+        InfantryWeapon rangeWeapon = infantry.getPrimaryWeapon();
+        if (infantry.getSecondaryWeapon() != null && infantry.getSecondaryN() > 1
+                && !infantry.getSecondaryWeapon().hasFlag(WeaponType.F_TAG)) {
+            rangeWeapon = infantry.getSecondaryWeapon();
+        }
+        boolean scuba = infantry.getMovementMode() == EntityMovementMode.INF_UMU
+                || infantry.getMovementMode() == EntityMovementMode.SUBMARINE;
+        hideElement(UW_LABEL, !scuba);
+        InfantryWeapon singleSecondary = (infantry.getSecondaryN() == 1)? infantry.getSecondaryWeapon() : null;
+        for (int j = 0; j <= 21; j++) {
```
sixlettervariables: The soldiers themselves are 1-based and ranges are 0-based?

neoancient: Yes, which could be a little confusing but it coincides with the text label. The soldiers and numbered 1-30 and the range table headers start at zero.

PR #580, opened 2020-04-19
Update the Maven Repo on source change

sixlettervariables:
This works like the one on the MM repository with some changes in paths.


sixlettervariables:
Same repo, MML is used by MHQ only from there.

PR #603, opened 2020-04-25
Strategic fuel text

### LINKED ISSUE #560: Fuel usage for JumpShips and space stations shouldn't be displayed as "Burn Days 1G"
Using MML 0.47.5 on OS X with Java build 1.8.0_51-b16, one of the ways that information on fuel usage is displayed for large aerospace units is "Burn Days 1G." This is correct for small craft, DropShips, and WarShips, but not for JumpShips or stations. These two units use station-keeping drives with only 0.1g of thrust.

The number displayed under "Burn Days 1G" is the correct amount of fuel for a burn-day at 0.1g, so it's really just the 1G part of the label that's wrong.

neoancient:
Updates the label text in the fuel view to reflect the difference in strategic fuel use between units with a transit drive (DropShip, WarShip) and those with station-keeping drives (JumpShip, space station). The units with station-keeping drive measure their consumption in burn-days at 0.1 G instead of 1.0 G and don't have a separate value for burn-days at max thrust.

Fixes #560.


src/megameklab/com/ui/view/FuelView.java
```diff
@@ -211,6 +215,15 @@ private void setFromAero(Aero aero) {
...
+                lblBurnDaysMaxText.setVisible(false);
+                lblBurnDaysMax.setVisible(false);
+            } else {
+                lblBurnDays1GText.setText(resourceMap.getString("FuelView.lblBurnDays1G.text")); //$NON-NLS-1$
+                lblBurnDaysMaxText.setVisible(true);
+                lblBurnDaysMax.setVisible(true);
+            }
         } else {
             panBurnDays.setVisible(false);
         }
```
Windchild292, line 229: Why is this not done as part of creating the panel?

PR #608, opened 2020-04-27
Show drone control count for ATAC and DTAC on record sheet

neoancient:
Adds ATAC and DTAC to DCCS and MASH as equipment with a virtual MiscType instance that expands capacity and needs to be consolidated on the record sheet. I realize this code is nearly identical to what is in the TRO summary template model (in fact that code was copied from here), but I can't think of an obvious place to put it for common access at this time.


neoancient:
Made obsolete by #610

PR #617, opened 2020-05-07
Equipment table sorting

### LINKED ISSUE #614: Change the list of equipment while sort by Crit section cause error
### Environment
Ver 0.47.5~0.47.6
Windows 10

### Description
Although I have seen the bug on 0.47.5, but I can't reproduce when it does so I didn't report it. But just now I can get how to do(and perhaps why it occurs). It only occurred on 0.47.5 and 0.47.6, but not 0.47.4.

You can cause the bug by following this step:

1. First, move to Equipment tab.
2. Choose any list as you want.
3. Click 'Crit' and sort by critical slots.
4. Then change the list to either **Physical Weapons**, **All Weapons** or **Other Equipment**.
5. See the difference? Now the list is down and you can't see anything correctly. Even if you click the other line and sort by the other element, it doesn't bring back to the normal state.

b1

I think that the reason is the item with 'Crit' value of 'variable'. Only the lists with such items are cause the error, such as Hatchet of Physical Weapon and Targeting Computer of Other Equipment.

Edit: But the weird thing is, 'Ton' line doesn't cause the same problem, despite there are also some items with Ton value of 'variable'. The bug only occurs when you sort by Crit.

neoancient:
This PR consolidates various Comparators used for sorting columns in the master equipment table on the equipment tab. Instead of having identical code in every equipment tab class there is a single copy of each Comparator in the table model class, with a lookup method that simplifies applying the correct Comparator to the column sorter. The sorting of columns consisting primarily of numeric values but sometimes Strings could throw a ClassCastException if the input was Integer or Double instead of String. This has been remedied so it could be applied to the BV and critical slots columns. I also added proper sorting for intro dates and lookup reference, because sooner or later somebody is going to report that they don't work as expected.

Fixes #614


src/megameklab/com/ui/Aero/tabs/EquipmentTab.java
```diff
@@ -176,15 +173,12 @@ public EquipmentTab(EntitySource eSource) {
...
-        equipmentSorter.setComparator(EquipmentTableModel.COL_RANGE, new WeaponRangeSorter());
-        equipmentSorter.setComparator(EquipmentTableModel.COL_COST, new FormattedNumberSorter());
-        equipmentSorter.setComparator(EquipmentTableModel.COL_TON, new FormattedNumberSorter());
+        equipmentSorter = new TableRowSorter<>(masterEquipmentList);
+        for (int col = 0; col < EquipmentTableModel.N_COL; col++) {
+            equipmentSorter.setComparator(col, masterEquipmentList.getSorter(col));
+        }
         masterEquipmentTable.setRowSorter(equipmentSorter);
-        ArrayList<RowSorter.SortKey> sortKeys = new ArrayList<RowSorter.SortKey>();
+        ArrayList<RowSorter.SortKey> sortKeys = new ArrayList<>();
```
Windchild292, line 181: Can this be a List?

src/megameklab/com/util/EquipmentTableModel.java
```diff
@@ -520,4 +549,95 @@ public Component getTableCellRendererComponent(JTable table,
...
+                // Simplest processing of Integer and Double
+                d1 = ((Number) o1).doubleValue();
+            } else if (o1.toString().endsWith("kg")) {
+                // Convert kg values to tons
+                d1 = Double.parseDouble(o1.toString().replace("kg", "").trim()) / 1000.0;
+            } else {
+                // Handle potentially commafied number
+                d1 = NumberFormat.getInstance().parse(o1.toString()).doubleValue();
+            }
+        } catch (NumberFormatException| ParseException ignore) {
```
Windchild292, line 571:
```suggestion
        } catch (NumberFormatException | ParseException ignored) {
```

src/megameklab/com/util/EquipmentTableModel.java
```diff
@@ -520,4 +549,95 @@ public Component getTableCellRendererComponent(JTable table,
...
+        }
+        try {
+            if (o2 instanceof Number) {
+                d2 = ((Number) o2).doubleValue();
+            } else if (o2.toString().endsWith("kg")) {
+                d2 = Double.parseDouble(o2.toString().replace("kg", "").trim()) / 1000.0;
+            } else {
+                d2 = NumberFormat.getInstance().parse(o2.toString()).doubleValue();
+            }
+        } catch (NumberFormatException|ParseException ignore) {
```
Windchild292, line 582:
```suggestion
        } catch (NumberFormatException | ParseException ignored) {
```

src/megameklab/com/ui/Aero/tabs/EquipmentTab.java
neoancient: Why? It's a local variable that is assigned a value once when it's instantiated.

neoancient:
1. What difference are you talking about?
2. Why?

src/megameklab/com/ui/Aero/tabs/EquipmentTab.java
Windchild292: Best practices are to use the List interface for the variable declaration unless you need something specific to ArrayList?

neoancient: It makes a difference in method parameters, where ArrayList would prevent passing any List as an argument that is not instanceof ArrayList, or when the variable might be assigned the results of an List expression, but neither is the case here. If I were writing this as new code I would declare it as List as habit, but as there is no problem with the way it was written and there's no reason to change it.

neoancient:
Initially I kept the same model of using the Comparators as inner classes, only I put them all in one place and made them static. They don't have any use outside this particular table so I saw no benefit to having them in their own files where they would just be clutter to sort through. Then I decided there's no need to create a new instance for every column, so I created a single instance of each.

PR #618, opened 2020-05-08
Fix text overlap in vehicle record sheet tables

### LINKED ISSUE #616: 0.47.6: Overlapping Text in Vehicle Record Sheet
Attached are an image of the issue, the PDF record sheet, and the zip file for the vehicle's blk.
image
[Cromwell CRM-4L.pdf](https://github.com/MegaMek/megameklab/files/4583139/Cromwell.CRM-4L.pdf)
[Cromwell CRM-4L.zip](https://github.com/MegaMek/megameklab/files/4583142/Cromwell.CRM-4L.zip)

neoancient:
I couldn't reproduce the issue, but going over the template generation code I found a bug that does not include fixed lengths for the segments of text in the last line of a paragraph when it includes normal and italic text. I am confident that this resolves the issue.

I also discovered several places where "inflict[ed]" is missing the L and "modifier" is missing the second I. I suspect that this is due to the way the TM PDF handles fl and fi ligatures, since I copied the text from the PDF (ironically) to avoid typos.

I'm not sure how to advise reviewing machine-generated files. There are some additional changes due to modifications of the generation code in the process of making the templates for conventional infantry and BA.

Fixes #616

### DIFF
--- a/data/images/recordsheets/submarine_dualturret_standard.svg
+++ b/data/images/recordsheets/submarine_dualturret_standard.svg
@@ -1303,7 +1303,7 @@
                                                                           ><text fill="#231f20" x="6.000" y="25.000" style="font-family:Eurostile;font-size:4.800px;font-weight:normal;font-style:normal" text-anchor="start"
                                                                           >for more information).</text
                                                                           ><text x="6.000" y="30.000" fill="#231f20" text-anchor="start" style="font-family:Eurostile;font-size:4.800px;font-weight:normal;font-style:normal" textLength="199.000" lengthAdjust="spacingAndGlyphs"
-                                                                          >A result of 12 on the Naval Combat Vehicles Hit Location Table may inflict critical hit</text
+                                                                          >A result of 12 on the Naval Combat Vehicles Hit Location Table may a inflict critical hit</text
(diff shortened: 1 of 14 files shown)


data/images/recordsheets/submarine_dualturret_standard.svg
```diff
@@ -1291,21 +1289,21 @@
...
                                                                           > </text
                                                                           ><text fill="#231f20" x="6.000" y="25.000" style="font-family:Eurostile;font-size:4.800px;font-weight:normal;font-style:normal" text-anchor="start"
                                                                           >for more information).</text
                                                                           ><text x="6.000" y="30.000" fill="#231f20" text-anchor="start" style="font-family:Eurostile;font-size:4.800px;font-weight:normal;font-style:normal" textLength="199.000" lengthAdjust="spacingAndGlyphs"
-                                                                          >A result of 12 on the Naval Combat Vehicles Hit Location Table may infict critical hit</text
+                                                                          >A result of 12 on the Naval Combat Vehicles Hit Location Table may inflict critical hit</text
```
Windchild292, line 1306: Maybe this should be may inflict a critical hit?

neoancient: Odd that didn't come up in my search for infict.

neoancient: Oh, maybe I should read better.

PR #631, opened 2020-05-21
Printing cleanup

neoancient:
This removes obsolete code and files that are no longer needed now that record sheet printing has been completely rewritten.


src/megameklab/com/printing/PrintRecordSheet.java
```diff
@@ -150,16 +150,6 @@ protected final Font getBoldFont(float size) {
...
-        // If the font is not installed, use system default sans
-        if (null == font) {
-            typeface = Font.SANS_SERIF;
-            font = Font.decode(typeface);
-        }
-        // If that doesn't work, get the default dialog font
-        if (null == font) {
-            font = Font.decode(null);
-            typeface = font.getName();
-        }
```
neoancient, line 162: If Font.decode doesn't find the requested typeface it returns the default font anyway.

src/megameklab/com/util/StringUtils.java
```diff
@@ -356,8 +260,8 @@ public static String getEquipmentInfo(Entity unit, Mounted mount) {
         return info.trim();
     }
 
-    public static String getEquipmentInfo(Aero unit, Mounted mount) {
-        String info = "";
+    public static String getAeroEquipmentInfo(Mounted mount) {
```
neoancient, line 263: The unit parameter was to distinguish this method from getEquipmentInfo(Entity, Mounted), but not actually used. The overloaded method required casting Entity to Aero to get the right one.

PR #639, opened 2020-06-02
New printing options: page size and color/monochrome

neoancient:
Adds two new options to record sheet printing.
1. The print dialog will honor the selection of paper size and margins. The image will be scaled to fit within the printable area. There are now two versions of each template, held in separate directories: one for US letter (the current size), and a new one that uses the ISO standard proportions for printing on A4 (or any other size within the standard). If the selected paper size does not fit one of these, the one closest to the aspect ratio is selected. The templates now have a single top-level group element that is used to rescale the sheet and set the margins by adding a matrix transformation.
2. The choice whether to use color printing or monochrome. The templates have monochrome copies of each color element, and one or the other is hidden depending on how the configuration option is set. The ids of the variable color items are embedded in the style attribute of the top-level group element.


src/megameklab/com/util/UnitPrintManager.java
```diff
@@ -258,9 +255,12 @@ private static File getExportFile(Frame parent, String suggestedFileName) {
...
     }
 
-    public static boolean exportUnits(List<Entity> units, File exportFile, boolean singlePrint) {
-        List<PrintRecordSheet> sheets = createSheets(units, singlePrint);
+    public static void exportUnits(List<Entity> units, File exportFile, boolean singlePrint) {
+        RecordSheetOptions options = new RecordSheetOptions();
+        List<PrintRecordSheet> sheets = createSheets(units, singlePrint, options);
         PageFormat pageFormat = new PageFormat();
+        // getPaper() returns a clone
+        pageFormat.setPaper(options.getPaperSize().createPaper());
```
Windchild292, line 263: This seems off to me.

neoancient: That comment is obsolete. Is it the disconnect between the comment and the code that's off?

Windchild292: It is

PR #643, opened 2020-06-10
Small unit record sheet fluff image

neoancient:
Sets the fluff image in the top right corner for infantry, BA, and protomech record sheets. These sheets can have multiple units and there is only space for a single image. In talking with @HammerGS we decided that the fluff image should be set if all the units on the page share the same chassis. Otherwise the default fluff image is used.

Also fixes some overlap issues in the BA and protomech templates I noticed during testing.


src/megameklab/com/printing/PrintSmallUnitSheet.java
```diff
@@ -109,4 +113,29 @@ protected String getRecordSheetTitle() {
         // Not used by composite sheet
         return "";
     }
+
+    private void drawFluffImage() {
+        if (entities.size() > 1) {
+            for (int i = 1; i < entities.size(); i++) {
+                if (!entities.get(i).getChassis().equals(entities.get(0).getChassis())) {
```
Windchild292, line 120: Wouldn't this be best compared the opposite way around?

neoancient: What difference does it make?

PR #648, opened 2020-06-14
Fix scaling and positioning of tables on vehicle record sheet.

neoancient:
My recent update allowing the user to set the paper size broke vehicle record sheets with a single unit on them. It throws an exception when trying to position the tables due to an illegal attribute, (transform where it should be translate) but it also doesn't take any custom margins into account. In the process of fixing it I simplified it by using the code already in PrintRecordSheet to handle the loading and sizing.

### DIFF
--- a/src/megameklab/com/printing/PrintRecordSheet.java
+++ b/src/megameklab/com/printing/PrintRecordSheet.java
@@ -297,7 +297,7 @@ public abstract class PrintRecordSheet implements Printable, IdConstants {
             createDocument(pageIndex, pageFormat);
             GraphicsNode node = build();
             node.paint(g2d);
-            /* Testing code that outputs the generated svg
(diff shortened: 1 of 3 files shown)


src/megameklab/com/printing/PrintRecordSheet.java
```diff
@@ -297,7 +297,7 @@ public int print(Graphics graphics, PageFormat pageFormat, int pageIndex) {
             createDocument(pageIndex, pageFormat);
             GraphicsNode node = build();
             node.paint(g2d);
-            /* Testing code that outputs the generated svg
+            /* Testing code that outputs the generated svg */
```
sixlettervariables, line 300: Is this test code still?

neoancient: Oops, missed that. I'll comment it back out.

PR #675, opened 2020-07-14
Allow Torso-Mounted (Industrial) Cockpit for IndustrialMech

binaryspica:
It is the part of MegaMek/MegaMek#2062, that allows to pick Torso-Mounted Cockpit for IndustrialMech, as well as also allows to pick Torso-Mounted Industrial Cockpit.


binaryspica:
Sigh. I don't get the problem, but was refer addTorsoMountedIndustrialCockpit() that is not on the master branch now causes the error?

Windchild292:
Different branch names

binaryspica:
Ah, I get it. Thanks.

PR #683, opened 2020-07-20
Remove jump jets/umu when required by year/tech level changes

### LINKED ISSUE #659: Jumping MP not updated if Improved Jump Jets are removed as a result of changing the production year of a Mek
Using MML 0.47.7 on OS X with Java build 1.8.0_51-b16, if you construct a unit with improved jump jets and set a Jump MP higher than the unit's Walk MP (as you should), and then change the construction year to before IJJs were developed, MML automatically converts IJJs to regular JJs. However, the unit keeps its (illegal) Jump MP. Also, the unit still has all of its original improved jump jets if you check the criticals.

If I just switch from IJJs to regular JJs without changing the year, the Jump MP is set to the unit's Walk MP. This seems more reasonable.

neoancient:
A change in the year or tech level can cause jump jets (IJJs in particular) or UMU to become invalid, but they are not being removed like other equipment.

Fixes #659

### DIFF
--- a/src/megameklab/com/ui/Mek/tabs/StructureTab.java
+++ b/src/megameklab/com/ui/Mek/tabs/StructureTab.java
@@ -659,12 +659,12 @@ public class StructureTab extends ITab implements MekBuildListener, ArmorAllocat
         }
         panChassis.refresh();
         panHeat.refresh();
-        panMovement.refresh();
         panArmor.refresh();
         panArmorAllocation.setFromEntity(getMech());
         panPatchwork.setFromEntity(getMech());
         refresh.refreshBuild();
         addAllListeners();
+        panMovement.refresh();
         refresh.refreshPreview();
     }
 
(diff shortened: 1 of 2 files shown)


src/megameklab/com/ui/Mek/tabs/StructureTab.java
```diff
@@ -659,12 +659,12 @@ public void updateTechLevel() {
         }
         panChassis.refresh();
         panHeat.refresh();
-        panMovement.refresh();
         panArmor.refresh();
         panArmorAllocation.setFromEntity(getMech());
         panPatchwork.setFromEntity(getMech());
         refresh.refreshBuild();
         addAllListeners();
+        panMovement.refresh();
```
Windchild292, line 667: Does this cause any issues being below refreshBuild and addAllListeners?

neoancient: The issue was caused in part by it being above that.

PR #691, opened 2020-07-22
Account for Artemis, PPC capacitor, etc. on large craft build tab

### LINKED ISSUE #679: LRM/SRM bays on large aerospace units do not reflect damage increase from Artemis IV
Using MML 0.47.7 on OS X with Java build 1.8.0_51-b16, the damage displayed for a missile bay on a large aerospace unit does not reflect the increase provided by Artemis IV. For example, a bay with two LRM 20s is listed as doing 24 points of damage, whether or not it contains an advanced FCS. According to the table on TM pp. 341-343, a pair of LRM 20s with Artemis IV should do 32 damage. I've seen the same issue with SRMs.

neoancient:
The total AV for a bay does not reflect the adjustment for linked equipment, nor does the tooltip for the bay or the individual weapon. The AV limit of 700 for a standard scale bay also doesn't take this into account.

Fixes #679

### DIFF
--- a/src/megameklab/com/ui/util/BayWeaponCriticalTree.java
+++ b/src/megameklab/com/ui/util/BayWeaponCriticalTree.java
@@ -1296,27 +1296,22 @@ public class BayWeaponCriticalTree extends JTree {
                 if (w.getType().hasFlag(WeaponType.F_PLASMA)) {
                     if (((WeaponType)w.getType()).getDamage() == WeaponType.DAMAGE_VARIABLE) {
                         av += 7;
-                    } else if (w.getType().hasFlag(WeaponType.F_ARTILLERY)) {
-                        av += ((WeaponType) w.getType()).getRackSize();
                     } else {
                         av += 13.5;
                     }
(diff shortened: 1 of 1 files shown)


src/megameklab/com/ui/util/BayWeaponCriticalTree.java
```diff
@@ -705,7 +708,41 @@ public String getTooltip() {
...
+                return 2;
+            }
+        } else if (linkedBy.hasFlag(MiscType.F_ARTEMIS_PROTO) && weapon.getRackSize() == 2) {
+            // The +1 cluster hit bonus only adds a missile hit for SRM2
+            return 2;
+        } else if (linkedBy.hasFlag(MiscType.F_PPC_CAPACITOR)) {
+            // PPC capacitors in weapon bays are treated as if always charged
+            return 5;
+        }
+        return 0;
```
Windchild292, line 743:
```suggestion
        }

        return 0;
```

PR #760, opened 2020-09-14
Edit small SV ammo capacity

### LINKED ISSUE #454: Can't add ammo for light and medium weapons to support vehicles
Using MML 0.47.4 on OS X, I can't add seem to add additional ammunition for light or medium weapons to a support vehicle. Maybe I've missed it?

neoancient:
Adds the ability to edit the amount of ammo carried for each weapon on a small support vehicle. The size column on the installed equipment table on the equipment tab does double duty.

Depends on MegaMek/megamek#2219

Closes #454


src/megameklab/com/util/UnitUtil.java
```diff
@@ -2524,7 +2529,6 @@ public static void updateLoadedUnit(Entity unit) {
     }
 
     public static void updateLoadedMech(Mech unit) {
-        UnitUtil.removeOneShotAmmo(unit);
```
NickAragua, line 2527: Do mechs not need one shot ammo removed any more? Or does the removeOneShotAmmo method get called somewhere else more generally?

NickAragua (commented):
One question, looks good otherwise.

src/megameklab/com/util/UnitUtil.java
neoancient: It now gets called in updateLoadedUnit(Entity), so it gets called for all units. This keeps it from showing up in the equipment list.

NickAragua (approved):

PR #786, opened 2020-10-03
Small fixed wing SV corrections

neoancient:
This is a companion PR to MegaMek/megamek#2293. It uses aero stats instead of ground stats for small fixed wing support vehicles on the master equipment table and on the record sheet. It also shows the number of shots of ammo for each weapon in the inventory, which is needed for all small support vehicles.

### DIFF
--- a/src/megameklab/com/util/EquipmentTableModel.java
+++ b/src/megameklab/com/util/EquipmentTableModel.java
@@ -341,7 +341,7 @@ public class EquipmentTableModel extends AbstractTableModel {
         } else if (col == COL_RANGE) {
             if (null != wtype) {
                 if (entity instanceof Aero) {
-                    switch (wtype.maxRange) {
+                    switch (wtype.getMaxRange(null)) {
                         case RangeType.RANGE_SHORT:
                             return "Short";
                         case RangeType.RANGE_MEDIUM:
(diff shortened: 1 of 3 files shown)


src/megameklab/com/util/EquipmentTableModel.java
```diff
@@ -341,7 +341,7 @@ public Object getValueAt(int row, int col) {
         } else if (col == COL_RANGE) {
             if (null != wtype) {
                 if (entity instanceof Aero) {
-                    switch (wtype.maxRange) {
+                    switch (wtype.getMaxRange(null)) {
```
neoancient, line 344: The null argument is for the mount, as is used for weapons like ATM and MML that vary the range depending on the type of ammo loaded. Passing null skips that check and just returns the calculated max range, which is identical to the previous behavior for heavy weapons, but now handles light and medium (e.g. infantry) weapons correctly.

src/megameklab/com/util/StringUtils.java
```diff
@@ -252,7 +250,7 @@ public static String getAeroEquipmentInfo(Mounted mount) {
         if (mount.getType() instanceof WeaponType) {
             WeaponType weapon = (WeaponType) mount.getType();
             if (weapon instanceof InfantryWeapon) {
-                info = Integer.toString(weapon.getDamage());
+                info = "";
```
neoancient, line 253: This is a copy-paste artifact from the ground unit method that adds this info to the damage field. For aerospace units the damage is part of the AV, and this is being appended to the weapon name.

NickAragua (approved):

PR #801, opened 2020-10-12
Printing progress bar and memory use improvement

### LINKED ISSUE #342: Incomplete queued print
When I try PDF printing a large queue of units with fluff images, either with a manual queue or a .MUL file, the program seems to give up after about 45 pages or so and the PDF is left unfinished. The problem doesn't seem to occur when printing without the images. To check if it was a problem with the Microsoft PDF virtual printer, I tested it with XPS Writer and CutePDF Writer and the problem still occurs. I also notice that the program slows down and uses a very large amount of RAM after attempting to print a large queue and the memory isn't freed until the program is closed.

neoancient:
This started as a fix for running out of memory on large print jobs, but adding the progress bar made it a lot easier to test.

1. Adds a non-modal popup with a progress bar that shows the progress of a job sent to the printer or exported to PDF. Anybody who doesn't like it can disable it in the configuration settings.
2. Moves the pdf export code into a background thread where it belongs instead of executing on the event dispatch thread.
3. Removes references to record sheet instances as they are processed so they can be garbage collected, and large jobs don't run out of memory. Before this change I could print about 40% of a 300-unit MUL. It now prints the entire document.

~~The memory usage change only works for export to PDF. I haven't been able to free up anything while processing a print job yet, but I'm going to leave this as a draft while I think about it. If I can't make it work, it will at least be possible to output to pdf and then send that to the printer.~~

The memory use changes now work for both printing and export.

Fixes #342.


src/megameklab/com/printing/RecordSheetTask.java
```diff
@@ -192,4 +187,41 @@ public Void doInBackground() throws Exception {
             return null;
         }
     }
+
+    /**
+     * Implementation of Pageable thar removes the record sheet objects as they are processed
```
sixlettervariables, line 192:
```suggestion
     * Implementation of Pageable that removes the record sheet objects as they are processed
```

sixlettervariables (approved):
One comment update needed, but it looks good!

src/megameklab/com/printing/RecordSheetTask.java
neoancient: I guess I'm a few weeks late for Talk Like a Pirate Day.

PR #805, opened 2020-10-15
Fix for biped superheavy structure pips not showing

neoancient:
I noticed while working on something else that the metadata in the biped templates that determines where to put the structure pips is displaced off the image below and to the right. Since the standard weight biped mechs uses a predetermined pattern of pips, it doesn't affect them. But the pip placement for superheavies is calculated and therefore doesn't show up. I suspect I messed something up in the xml source when added a field for structure type.


neoancient:
I know these SVG updates aren't very helpful. Here's what the generated SVG document looks like with the misplaced pips.

Screenshot from 2020-10-14 17-26-24

sixlettervariables (approved):

Issue #176, opened 2018-08-05
Include Optional Quick Ref Charts


pmjdebruijn:
Currently MegaMekLab produces nice printed record sheets.

It does not seem to have an option to include some common quick reference charts onto the sheets as the defunct Solaris Skunk Werks had:

https://raw.githubusercontent.com/Solaris-Skunk-Werks/solarisskunkwerks/master/data/rs/Charts.png

HammerGS:
Going to close this, but please it to the open ticket here - https://github.com/MegaMek/megameklab/issues/37

pmjdebruijn:
Would it be possible to reopen?

I've made a first attempt at creating a border (consider it nothing more than a proof of concept):
https://encrypted.pcode.nl/files/temp/btsheetchartborder.svgz

Which results in (manually composited in Inkscape) something like:
https://encrypted.pcode.nl/files/temp/btsheetchart.pdf

I've tried to keep typefaces and sizes roughly similar to what the scaled sheet should have.

Is this something either you or @neoancient could give some feedback on?

HammerGS:
The issue is still open.

neoancient:
It's something that's been kicked around for a while. I did a little work on this primarily as proof of concept but haven't had the time to devote to it. The record sheets for mechs are an SVG template which contain metadata that tells MML where to fill in data. The idea is to provide an alternate template file that has the record sheet part of the image scaled down to provide space for tables at the bottom and side.

HammerGS:
As an extension of this, any changes to RS we would have to get CGL's feedback on it, as they use MML to generate RS for the books.

pmjdebruijn:
Presumably it should always be an option, as scaling the standard chart to make room for the quick ref tables does make for smaller type. So folks with less than optimal eye-sight would not appreciate this.

BTW, I've played a bit with modifying biped_mech_default.svg, however when I make big adjustments like scaling, it all breaks down, for example:

https://encrypted.pcode.nl/files/temp/Catapult_CPLT-C1__DF-job_1.pdf

Now I did add the quick ref tables sloppily myself, so don't mind that. But all the original sheet content, was only scaled and moved a little (all with everything selected at once).

Is there anything in particular I need to be mindful off when trying to manipulate biped_mech_default.svg?

HammerGS:
yes, this would need to be an option.  We don't have a lot of options to change the RS without impacting CGL's ability to produce RS for product.

Internally we've had conversations about adding Tables and Charts back.  

[...]

PR #811, opened 2020-10-26
Add optional reference tables for vehicle record sheets

neoancient:
Adds reference sheets to vehicle record sheets when the option is selected. The tables are customized for the type of vehicle. The reference tables option is assumed to imply a single unit per sheet for tanks so you get the hit location, critical, and motive system damage tables.

Here is a sample that includes an assortment of motive types, and also includes a support vehicle without the offroad chassis modification and a tank with a limited amphibious chassis modification to illustrate the changes those cause in the movement cost table.
[vee_test.pdf](https://github.com/MegaMek/megameklab/files/5436391/vee_test.pdf)

I also noticed a small overlap in the two halves of the base sheet, which is made more obvious when scaled down to make room for the tables. The bottom half was not taking the top margin in account when calculating the center of the sheet.

Partial completion of #176.


src/megameklab/com/printing/reference/DrivingSkillRollMods.java
```diff
@@ -0,0 +1,80 @@
...
+            addRow(bundle.getString("sideslipMovement"), "", "", "");
+        } else {
+            addRow("", bundle.getString("flankingMovement"), "",
+                    bundle.getString("possibleSkid"));
+            addRow("", bundle.getString("afterFacingChange"), "", "");
+            addRow("", bundle.getString("onPavement"), "", "");
+            addRow(bundle.getString("skiddingMovement"), "", "", "");
+        }
+        addRow("", bundle.getString("hexesMovedInTurn"), "", "");
+        addRow("", "", "0-2", "-1");
```
NickAragua, line 59: This seems like it's off by 1, is this not "to-hit" modifiers for when the unit is being shot at?

NickAragua (commented):
Looks fine, one question.

src/megameklab/com/printing/reference/DrivingSkillRollMods.java
neoancient: This is the driving skill check to avoid skidding when moving at flanking speed on pavement after a facing change.

NickAragua (approved):

PR #820, opened 2020-11-06
Small unit tables

### LINKED ISSUE #176: Include Optional Quick Ref Charts

neoancient:
This is the final batch of optional reference tables for record sheets, dealing with the infantry, battle armor, and protomech sheets. Like the larger units, these save space by omitting data not relative to the unit on the sheet. This is a bit more involved when there are 3-5 units to cover, which may have to be distinguished. One change is the addition of a column on the movement cost table for prohibited motive types, which is used if there are any mechanized infantry on the sheet Another (slightly awkward) adjustment is that most conventional infantry require a lot of columns on the cluster hits table, so that is provisionally moved to the bottom in place of the movement record. Also when the reference sheets are used, the number of conventional infantry units per sheet is limited to three instead of the usual four to leave room for the infantry damage tables.

Conventional infantry sample:
[infantry_test.pdf](https://github.com/MegaMek/megameklab/files/5499023/infantry_test.pdf)

Battle armor sample:
[ba_test.pdf](https://github.com/MegaMek/megameklab/files/5499025/ba_test.pdf)

Protomech sample:
[protomech_test.pdf](https://github.com/MegaMek/megameklab/files/5499027/protomech_test.pdf)

Closes #176


NickAragua:
Question. Why do the protomech hit location tables have gaps in them? e.g. 2 = main gun, 4 = right arm. What happened to 3 and 11? Are those just "roll again"?

neoancient:
Those are near miss, except in the case of a glider protomech (see the Svartalfa). That has nothing to do with the PR. This PR does include the special damage allocation table, which is used when damage can't miss, such as an area effect weapon, a fall, or a collision with a skidding unit.

NickAragua (approved):

PR #828, opened 2020-11-11
Upgrade to Gradle 6.7

sixlettervariables:
This upgrades to Gradle 6.7 and resolves the deprecations noted during builds. This also removes the unused (and likely unfixable) Mac OS X bundle.

When including MegaMekLab under Gradle 6+ you'll need to give the build a name, like in your `settings_local.gradle`:
```groovy
includeBuild('../megameklab') {
    name = 'mml-composite'
}
```
No two `GradleBuild` tasks or includes can share the same name. Under Gradle 6.3+ it uses the directory name by default, leading to these issues.


Windchild292 (approved):
For immediate merge post-stable

PR #839, opened 2020-12-02
MegaMek Issue #2465: Ignore crew effects in MP calculations

sixlettervariables:
This is the MML companion to https://github.com/MegaMek/megamek/issues/2465

If we decide to revert the changes to getWalkMP/getRunMP, this PR will be obsolete in most respects.

However, I did notice that Tanks and Support Vees had some inconsistency in how they asked the entity to calculate their Run MP. I don't think this matters in practice for MML, as there is no IGame object (I do not believe), but would be worth cleaning up.


sixlettervariables:
Insert: "there has got to be a better way"

PR #893, opened 2021-03-13
Do not allow unit selector to become rooted

sixlettervariables:
While debugging memory issues with MML on the new PDF exports branch I noticed N rooted copies of MegaMekLabUnitSelectorDialog due to cached visuals, where N equals the number of times I opened the unit selector. This refactors the launching of the unit selector dialog to ensure the resources are released. They are no longer part of the rooted list.

Before:
image

After:
image

### DIFF
--- a/src/megameklab/com/ui/StartupGUI.java
+++ b/src/megameklab/com/ui/StartupGUI.java
@@ -36,7 +36,6 @@ import javax.swing.JLabel;
 import javax.swing.JOptionPane;
 import javax.swing.UIManager;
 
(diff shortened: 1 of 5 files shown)


sixlettervariables:
Diving a bit further, I may have to do some more work. Looks like reference cycles between `UnitSelectorDialog` and the `JFrame` also cause an issue here.

Windchild292:
Closing, opened https://github.com/MegaMek/megamek/issues/3264 to replace this.

PR #918, opened 2021-06-03
Show unused tonnage in the status bar.

### LINKED ISSUE #813: minor feature request - show available tonnage directly
One thing that I always hated about MML compared to Solaris SkunkWerks is the fact that MML only shows how much tonnage has been used out of the overall tonnage, while SSW (also) directly shows the tonnage available (positive or negative), which makes it much easier to figure out ideal weight brackets etc., while in MML you always have to do the math in your head instead of just looking at the relevant number, which is sort of annoying.
I don't think it's difficult to include this improvement, but it would seriously benefit various situations tinkering around in MML.

neoancient:
New format:
Screenshot from 2021-06-03 11-16-55

Besides adding the remainder, BA and small support vehicles are now displayed in kg (protomechs were already displaying in kg), and large craft tonnages are commified (using locale-appropriate grouping character).

Closes #813

### DIFF
--- a/src/megameklab/com/ui/Aero/StatusBar.java
+++ b/src/megameklab/com/ui/Aero/StatusBar.java
@@ -118,7 +118,7 @@ public class StatusBar extends ITab {
         }
         heatSink.setVisible(getAero().getEntityType() == Entity.ETYPE_AERO);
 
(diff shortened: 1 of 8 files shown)


sixlettervariables (approved):
Neat use of the format string for the correct comma, did not know that was available in Java.

PR #923, opened 2021-06-07
Fix Beagle range on record sheet

### LINKED ISSUE #873: Beagle Active Probe range on record sheet
Record sheets currently display the Beagle Active Probe's range as 3 hexes; as per the BMM and TW, this should be changed to 4 hexes.

neoancient:
The max range of a BAP is 4 hexes.

Fixes #873


NickAragua (approved):

HammerGS (approved):
Seems like a simple change.

PR #924, opened 2021-06-07
Fix reengineered laser type on record sheet

### LINKED ISSUE #884: Reengineered laser damage type
v.47.17

As per the BMM p.101, re-lasers, on printed record sheets, should have their [DE] damage type changed to [P].

neoancient:
As a type of pulse laser, the reengineered lasers should have a damage type of [P] instead of [DE].

Fixes #884

### DIFF
--- a/src/megameklab/com/util/StringUtils.java
+++ b/src/megameklab/com/util/StringUtils.java
@@ -181,7 +181,7 @@ public class StringUtils {
                 }
                 if (UnitUtil.isAMS(weapon) || (weapon.hasFlag(WeaponType.F_B_POD))) {
                     info += "PB,";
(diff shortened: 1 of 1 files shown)


NickAragua (commented):
Wouldn't we want to address this by giving Re-Engineered weapons the F_PULSE flag instead?

HammerGS (approved):
Looks right to me.

neoancient:
> Wouldn't we want to address this by giving Re-Engineered weapons the F_PULSE flag instead?

I think you're right. I just looked through the code to see how F_PULSE is used and found the one place where it would make a difference is that it would disallow aimed shots with a targeting computer, which seems like it should apply to RELs also.

neoancient:
Superceded by MegaMek/megamek#2921

PR #929, opened 2021-06-12
Don't show protomech-only ammo on equipment tab for non-protomechs

### LINKED ISSUE #897: Additional peculiar entry for Clan Machine-Gun Ammo?
v0.48

There appears to be an additional Darkage-Era Machine Gun ammo (full ton) entry with a 0 c-bill cost and that exports as `Clan Machine Gun Ammo - Proto`

The Puma I and Gladiator T .mtfs from the ILClan Rec-guides contain this potentially duplicate ammo type.  

Screen Shot 2021-03-16 at 3 09 08 PM

neoancient:
This affects Clan machine guns.

Fixes #897

### DIFF
--- a/src/megameklab/com/util/UnitUtil.java
+++ b/src/megameklab/com/util/UnitUtil.java
@@ -4006,6 +4006,9 @@ public class UnitUtil {
                 && !atype.canAeroUse()) {
             return false;
         }
+        if (atype.hasFlag(AmmoType.F_PROTOMECH) && !(unit instanceof Protomech)) {
+            return false;
+        }
 
         for (Mounted m : unit.getTotalWeaponList()) {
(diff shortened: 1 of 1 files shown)


src/megameklab/com/util/UnitUtil.java
```diff
@@ -4006,6 +4006,9 @@ public static boolean canUseAmmo(Entity unit, AmmoType atype, boolean includeOne
                 && !atype.canAeroUse()) {
             return false;
         }
+        if (atype.hasFlag(AmmoType.F_PROTOMECH) && !(unit instanceof Protomech)) {
```
Windchild292, line 4009: Why not do an else if?

Windchild292 (approved):

PR #977, opened 2021-12-12
Log4j 2 Swapover

Windchild292:
This swaps MegaMekLab to Log4j 2 and by doing so improves our logging implementation in MegaMekLab.

The MegaMekLab::showInfo method has been standardized with MegaMek and MekHQ, and made public so that MekHQ can call it.

I chose to isolate all logging in MegaMekLab to the MegaMekLab.log file outside of legacy exceptions, as this makes it easier to bugfix.


SJuliez (approved):
OK

PR #993, opened 2022-01-02
Logging: Default Global Exception Handler and Standardized Legacy Logging

Windchild292:
MegaMekLab side of https://github.com/MegaMek/megamek/pull/3346

### DIFF
--- a/src/megameklab/com/MegaMekLab.java
+++ b/src/megameklab/com/MegaMekLab.java
@@ -1,5 +1,6 @@
 /*
- * MegaMekLab - Copyright (C) 2008
+ * MegaMekLab
+ * Copyright (c) 2008-2022 - The MegaMek Team. All Rights Reserved.
  *
  * Original author - jtighe (torren@users.sourceforge.net)
  *
@@ -28,7 +29,8 @@ import org.apache.logging.log4j.LogManager;
 
 import javax.swing.*;
 import java.awt.*;
-import java.io.*;
+import java.io.File;
+import java.io.IOException;
 import java.text.NumberFormat;
 import java.time.LocalDate;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


SJuliez:
Interesting. Does this affect the AWT event thread? I'd expect exceptions to mostly happen there.

Windchild292:
I tested using AWT, for exactly that reason. Plus it's a thread we don't directly control, which ensures that it holds across the system.

With AWT Exceptions it won't do the graphical display, but does properly log the exception to the primary log file (MegaMekLab.log for MML).

Windchild292:
image

Windchild292:
I will be watching the legacy log though, to ensure everything is caught. Once I'm done converting the remaining legacy logging to the log manager I'll remove the clearing functionality, and hopefully we will see an empty log from our constant dev version players. If those stay empty for a few releases, I'll know we can (finally) remove the old redirect.

SJuliez (approved):

SJuliez:
Just for the record, I gave it a quick test by changing line 48 in MegaMekLabMainUI to
menubarcreator = new MenuBar(null);
(replace "this" with "null"). It just closed down MML without any dialog popping up. But the exception is still in the log file, so all in all I experienced no change.

Windchild292:
It does log on the proper log file though, not the legacy one... right? If so, then it's still an improvement for that case

SJuliez:
It appeared in megameklab.log. Is megameklablog.txt the legacy one then? (Sorry Im a bit oblivious there, I tend to comment out the logging to have exceptions in the console)

Windchild292:
Legacy.log will be the new legacy logger, which is the console redirect. LogManager writes to MegaMekLab.log, so it's an improvement above the current and will help us remove our ancient logging redirect.

Windchild292:
I've also sent you a Log4j2.xml file that will let you redirect to console going forwards.

PR #1018, opened 2022-01-27
Improved Initial Log Message

Windchild292:
This is the MML implementation for https://github.com/MegaMek/megamek/pull/3412

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,6 @@
 /logs/
 /.settings/
 /classes
-timestamp
 /*.ps
 /MegaMekLab.jar
 .DS_Store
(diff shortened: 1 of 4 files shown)


SJuliez (approved):
Seeing the project_name constant I was wondering if there is a way to have static access to which program is the main programming running right now (so MML in MHQ could still know that this is MHQ running and would know if it's standalone MML). Is there a direct way to this or can we implement a SuiteConstants "constant" that is set to MM/MML/MHQ in the respective main()?

Windchild292:
That's my next goal for the suite changes, namely as part of standardizing the about dialogs. I wanted to handle the logging improvements to make debugging logs easier faster, but that's the goal for the whole design.

PR #1027, opened 2022-02-13
Added Coolant Pod to Other category check in EquipmentDatabaseCategory

### LINKED ISSUE #1026: Coolant pod not being shown in equipment tab
### Environment
MekLab standalone, latest version from Git (0.49.7-snapshot#e7d7844)
Windows 10, Java 11

### Description
In some recent commit the **Coolant Pod** disappeared from the "_Other_" section in the _Equipment_ tab. This means it is impossible to add it to a design as far as I can tell.

From looking at the source code the issue seems to be the fact Coolant Pods are implemented as _AmmoType_ due to their chance to explode.
Looking at the definitions of the categories in the class `EquipmentDatabaseCategory` shows this:
- The "_Ammo_" category is defined to only show ammunition the Mech design can fire, which obviously will never be true for a **Coolant Pod**.
- The "_Other_" category is defined as having to be equipment either of `MiscType `or `TAGWeapon`.

It seems the easiest solutions would be to add the code 
`eq instanceof AmmoType && ((AmmoType)eq).getAmmoType() == AmmoType.T_COOLANT_POD))`
at the end of the `OTHER ` enum's  lambda category definition (row 85), allowing Coolant Pod to appear again in the _Other_ category listing.

I added a PR with my proposed change as an example (#1027 )

redbatz:
fixes #1026

### DIFF
--- a/src/megameklab/com/ui/util/EquipmentDatabaseCategory.java
+++ b/src/megameklab/com/ui/util/EquipmentDatabaseCategory.java
@@ -82,7 +82,7 @@ public enum EquipmentDatabaseCategory {
             && !eq.hasFlag(F_SPONSON_TURRET)
             && !eq.hasFlag(F_PINTLE_TURRET))
             || (eq instanceof TAGWeapon)
-            || (eq instanceof AmmoType && ((AmmoType)eq).getAmmoType() == AmmoType.T_COOLANT_POD)),
+            || ((eq instanceof AmmoType) && (((AmmoType) eq).getAmmoType() == AmmoType.T_COOLANT_POD))),
 
     AP ("Anti-Personnel",
             (eq, en) -> UnitUtil.isBattleArmorAPWeapon(eq),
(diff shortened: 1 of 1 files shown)


src/megameklab/com/ui/util/EquipmentDatabaseCategory.java
```diff
@@ -81,7 +81,8 @@
             && !(eq.hasFlag(F_PARTIAL_WING) && en.hasETypeFlag(Entity.ETYPE_PROTOMECH))
             && !eq.hasFlag(F_SPONSON_TURRET)
             && !eq.hasFlag(F_PINTLE_TURRET))
-            || (eq instanceof TAGWeapon)),
+            || (eq instanceof TAGWeapon)
+            || (eq instanceof AmmoType && ((AmmoType)eq).getAmmoType() == AmmoType.T_COOLANT_POD)),
```
Windchild292, line 85:
```suggestion
            || ((eq instanceof AmmoType) && (((AmmoType) eq).getAmmoType() == AmmoType.T_COOLANT_POD))),
```

Windchild292 (commented):
Changes look good, just a few style bits first.

Windchild292 (approved):

Windchild292:
Thanks for the bugfix!

PR #1079, opened 2022-03-28
Fixing BuildView createLabel Missing Tooltips

Windchild292:
For Review:
The first commit is a refactor, the second removes some useless setEditable calls, the third adds tooltips to a bunch of labels that should have had them already.


SJuliez (approved):
While the labels had no TT, the text fields/combos did have them. But there's no disadvantage to labels also having TTs.

PR #1090, opened 2022-04-20
MML Menu Bar Rework

Windchild292:
This reworks MML's menu bar to make it more similar to MHQ's while expanding the mnemonic setup.


SJuliez:
Fired this up. A few points I noticed:
- There are two spacers between Options and Exit in the file menu
- The Ctrl-P shortcut for print current seems to have been lost. This was an obvious one and IMO worth preserving
- I'm not sure if some of the shortcuts are as intended: Alt-U will bring me to the unit menu regardless of where Im currently but the export/pdf menu lists Alt-U and others as a shortcut. pressing U when in there works (=mnemonic), but Alt-U never does. 
- Ctrl-L doesnt load. If I had to guess, it probably didnt before this PR either. Maybe worth bringing into the present by adopting this as another obvious shortcut.

Windchild292:
- Fixed
- Fixed
- This was a conscious decision I made when working on MekHQ's menu bar, and continued to use here, as keeping the menu bar consistent between MML and MHQ makes the most sense (I almost entirely agree with your MM decision for it). It's often hard to tell what the current character is in the text, if it even appears in the actual sentence. This shows the character for it properly, including the held alt required. Sadly, I didn't find a way to just have it show the entire keypath instead, cause that would be the best way to handle it.
- Fixed

SJuliez (approved):

PR #1141, opened 2022-07-04
Add missing columns to BA cluster reference table

neoancient:
The cluster hits optional reference tables only include the cluster hits columns for missile weapons for a full-strength BA squad. As troopers are eliminated other columns are needed.

### DIFF
--- a/megameklab/src/megameklab/printing/reference/ClusterHitsTable.java
+++ b/megameklab/src/megameklab/printing/reference/ClusterHitsTable.java
@@ -74,7 +74,9 @@ public class ClusterHitsTable extends ReferenceTable {
             if (entity instanceof BattleArmor) {
                 for (Mounted mounted : entity.getIndividualWeaponList()) {
                     if (mounted.getType() instanceof MissileWeapon) {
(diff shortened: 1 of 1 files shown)


Windchild292 (approved):

megameklab/src/megameklab/printing/reference/ClusterHitsTable.java
```diff
@@ -74,7 +74,9 @@ private void calculateClusterSizes(Entity entity) {
             if (entity instanceof BattleArmor) {
                 for (Mounted mounted : entity.getIndividualWeaponList()) {
                     if (mounted.getType() instanceof MissileWeapon) {
-                        clusterSizes.add(Math.min(40, size * ((MissileWeapon) mounted.getType()).getRackSize()));
+                        for (int i = 1; i <= size; i++) {
```
Windchild292, line 77: Just being very careful by double checking this is supposed to start at 1? It reads like it is, but better to be safe.

neoancient: It's the number of active troopers. I changed the loop variable name to make it more evident.

PR #1270, opened 2023-06-11
Eclipse IDE multiple Module error fix - take 2

HoneySkull:
Second take at fixing the Eclipse IDE Multiple Module error in a way that doesn't also break Scaled Vector Graphics printing.  This attempt excludes the offending `xml-apis` group but then adds the `xml-apis-ext` as runtime only which ensures it is included in the final build but doesn't trip up Eclipse IDE.

With this fix I'm able successfully import the megameklab gradle project into the Eclipse IDE while still able to print and export-to-pdf.  Further test is needed to ensure no other edge cases trigger "class not found" errors.

### DIFF
--- a/megameklab/build.gradle
+++ b/megameklab/build.gradle
@@ -36,27 +36,26 @@ dependencies {
         // We don't need the python and javascript engine taking up space
         exclude group: 'org.python', module: 'jython'
         exclude group: 'org.mozilla', module: 'rhino'
-        // Uncomment to work around Eclipse IDE Multiple Dependency errors. 
-        //exclude group: 'xml-apis'
+        exclude group: 'xml-apis'
     }
     implementation ('org.apache.xmlgraphics:batik-codec:1.14') {
-        // Uncomment to work around Eclipse IDE Multiple Dependency errors. 
-        //exclude group: 'xml-apis'
+        exclude group: 'xml-apis'
     }
     implementation ('org.apache.xmlgraphics:batik-dom:1.14') {
-        // Uncomment to work around Eclipse IDE Multiple Dependency errors. 
-        //exclude group: 'xml-apis'
+        exclude group: 'xml-apis'
     }
     implementation 'org.apache.xmlgraphics:batik-rasterizer:1.14'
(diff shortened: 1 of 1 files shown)


HammerGS:
Due to some real life commitments I won't be able to test this until later this week. But will provide feedback when its done.

HammerGS:
Hi, quick update on this. I was middle of moving and overly optimistic on my time to test. I'm now set back up and will test this weekend and get some feedback.

HammerGS:
Been testing this both running from the repo, and with local builds and it seems to work with no issues.

HoneySkull:
@HammerGS I've been testing with this as well with all configuration options, reference tables, and heat scales enable.   I'm trying to go through methodically all the impacted classes flagged in Eclipse IDE with the multiple module dependency error and hit each functional point.  I haven't completed my testing yet due to RL - but should be complete soon.

HoneySkull:
I'm not able to load Gun Emplacements due to invalid unit warnings.  But this seems to be a problem in the nighties as well.

HammerGS:
> I'm not able to load Gun Emplacements due to invalid unit warnings. But this seems to be a problem in the nighties as well.

Gun Emplacements are all hand made and unprintable. They are a major hack.

HammerGS:
@HoneySkull Chatting with Juliez we are comfortable merging this and having it hit nightlies and test it hard from the nightlies. 

Not sure if anything needs to happen on the MekHQ or MegaMek side.

HoneySkull:
Megamek requires no changes.   MekHQ will need similar Gradle updates in order for the MekHQ project to be successfully imported into EclipseIDE.  I'll start on the MekHQ update.

SJuliez (approved):

SJuliez:
@HoneySkull Is there anything left to add here or can we convert un-draft it?

HoneySkull:
Ready as is.

HammerGS:
Thanks. Where you going to also look at MekHQ?

HoneySkull:
@HammerGS - Yes - now that this is working in Megameklab, I will move to MekHQ to make similar adjustments.

PR #1293, opened 2023-07-30
Fix #1292: let File -> Exit option actually exit MegaMekLab

### LINKED ISSUE #1292: [GUI] File menu "exit" modal does not actually exit the program, while the corner [X] modal does.
### Description
The exit prompt modal that appears when user selects `File -> Exit`  does not exit MegaMekLab.
The exit prompt modal that appears when a user clicks on the [X] button in the top-right corner of the frame _does_ exit.

Expectation:
- Both methods instantiate the same modal with the same code and functionality.
- Answering "Yes" or "No" in the Modal either saves, or does not, followed by exiting the program.

Actual:
- They don't function the same.
- Neither option in the first modal exits.

This affects both custom and cache-only units.
This happens whether the user selects "Yes" or "No", although the Save functionality _does_ work in both modals.

### Files

[megamek.log](https://github.com/MegaMek/megameklab/files/12208209/megamek.log)

#### CLI debug logging for MML showing `File -> Exit` modal saving but not exiting, then `[X]` modal saving _and_ exiting.

[20230729_1605_mml_exit_issue.log](https://github.com/MegaMek/megameklab/files/12208210/20230729_1605_mml_exit_issue.log)

Sleet01:
See issue for more info.

I couldn't find a nice, clean ternary way to code this so I added a small function.


Sleet01:
Closing and resubmitting PRs as separate branches.

Issue #1156, opened 2022-08-12
RFE make 'auto fill unhittables' and 'auto compact' saveable  option


binaryspica:
### Environment
Ver 0.49.8

### Description
This makes me headache all the time when I design a mech. It is always active when I start one and I have to disable both of them manually each time. Isn't it need to be active all the time? I think that it should be turned off by default, for it puts the unhittables through where I want to put the equips and forces to change the location of the equips when I remove it on the location. It is more a problem than a handy addon.

HammerGS:
The current code follows the guidelines we’ve been provided to determine crit placement. 

We need more details to determine if it’s a bug or simply a feature that isn’t liked.

foij00:
Seconding request.  MML now autofills unhittables before you even have a chance to place other crit items, like weapons, heat sinks, etc.  It forces me to turn the autofill off before adding anything from the equipment tab.

gsparks3:
> it puts the unhittables through where I want to put the equips

Just to check ... you're aware that when this change was made, it was also set up so that dropping an item "into" a slot containing a roll-again crit will automatically move that crit somewhere else, right? You can just drop your equipment wherever you want it, and the "unhittables" will automatically move out of the way as needed.

foij00:
Thanks for the info, that helps.  However, I most frequently use MML to build canon mechs from official record sheets (before those units get uploaded to Megamek in a dev release), and I'd prefer not to 'chase' the unhittables around the mech to drag them to the correct position as listed on the canon record sheet.  I also figured out that turning off autofill and then using the reset button gets me what I want as well.

binaryspica:
Yes it doesn't allow me to place it where I want. Also just fill the space that what I want by default is very annoying as well. Why it is automatically and cannot be disabled by default? If you really want to fill it automatically you can just hit 'Fill' button once. But for now it makes me to manually click 'Auto Fill Unhittables' and 'Auto Compact' manually to disable it each time I edit a mech.

HammerGS:
While I use these and they do take some getting used too. I’m seeing continuous comments and requests around these. 

Could we look into making them an option that is saved?

Sleet01:
@HammerGS I'd like to take a stab at this, targeting 0.49.15, if that's okay?

HammerGS:
> @HammerGS I'd like to take a stab at this, targeting 0.49.15, if that's okay?

Sounds good.

Sleet01:
[...]

PR #1296, opened 2023-08-03
RFE #1156: make 'auto fill unhittables' and 'auto compact' saveable o…

Sleet01:
…ption

Problem:

  Some users want to turn off Auto Fill Unhittables and Auto Compact by
default, so that they can assign all criticals directly without having to reset or move auto-assigned items.

Solution:

  1. Add two new checkboxes to Misc Settings Panel for these options. A. Both are enabled by default. B. Both have suitable text and tooltips describing functionality.
  2. Have BMBuildTab read these CConfig values when instantiating the relevant buttons.

Testing:

  1. Tested new options: Checkboxes can be set/unset, saved, and loaded correctly.
  2. Tested with new Meks; default HS and added unhittable items are not filled or compacted.
  3. Tested with loaded mechs; changed items are not filled or compacted.

### DIFF
--- a/megameklab/resources/megameklab/resources/Dialogs.properties
+++ b/megameklab/resources/megameklab/resources/Dialogs.properties
@@ -50,6 +50,10 @@ ConfigurationDialog.chkSummaryFormatTRO.text=Use TRO format for text export
 ConfigurationDialog.chkSummaryFormatTRO.tooltip=When checked, text exports are formatted in technical readout style, otherwise as a traditional MegaMek unit summary.
 ConfigurationDialog.chkSkipSavePrompts.text=Disable save prompts. Use at your own risk!
 ConfigurationDialog.chkSkipSavePrompts.tooltip=When checked, no safety dialogs warning about losing changes when switching unit or unit type will be shown.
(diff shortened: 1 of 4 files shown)


SJuliez:
Oooooh as it has to happen at some point, I did the same but have not yet made a PR of it. Now what I did differently is that the offending buttons themselves simply save their state whenever clicked. So when disabling one of them, it'll stay disabled when loading the next mek. Seems simpler than having it in the options? I'll make a PR shortly, then we can see which one to use :)

Sleet01:
> Oooooh as it has to happen at some point, I did the same but have not yet made a PR of it. Now what I did differently is that the offending buttons themselves simply save their state whenever clicked. So when disabling one of them, it'll stay disabled when loading the next mek. Seems simpler than having it in the options? I'll make a PR shortly, then we can see which one to use :)

The specific request in the RFE was for a saveable setting that would persist, so that's what I did.

SJuliez:
My solution also persists. I did the same in Cconfig.

HammerGS:
@Sleet01 thansk for the PR much appreciated. 

But to me something that automatically remembers a setting is a better option.

Sleet01:
Sounds like the other PR is preferred; closing.

PR #1331, opened 2023-11-10
Beast mounted infantry

neoancient:
Adds beast-mounted as a possible infantry motive type. When beast-mounted is selected it defaults to horse to avoid the problems of having an undefined and untracked mount. There is a new mount tab that has two views. One allows selection of a predefined mount and the other allows the user to define a custom mount. The infantry record sheet adds notes as appropriate to the mount.

Depends on MegaMek/megamek #4883


Windchild292 (changes requested):
Two typos, otherwise this looks good

PR #1420, opened 2024-02-08
Allow construction of BA exoskeleton before 2710

### LINKED ISSUE #1203: Unable to assign year earlier than 2710 when creating Exoskeleton
When designing exoskeletons and you set the date, if that date is below 2710 it will revert to 2710 even though the intro date for exoskeletons is very early according to the rules (common 2200 IO 1e p50).

I _think_ it does this because the PA(L) suit has an intro date of 2710, which is part of this particular dropdown entry. 

This appears to be a rather old issue.

neoancient:
Use the exoskeleton tech advancement when exoskeleton is checked.

This required a workaround for there being no legal armor between 2100 and ~2710. So the No Armor type is used when no other armor is available, and the armor points spinner is set to zero and disabled.

Also made some UI improvements to clear the exoskeleton checks when the weight class is changed to something heavier, and use the correct intro dates for IS/Clan tech base.

Fixes #1203


HammerGS (approved):
Tested and working.

Issue #1429, opened 2024-02-17
NPE: Cannot invoke "megamek.common.Mounted.getLinked()" because "weapon" is null


HammerGS:
### Environment
What version of MekLab does your issue apply to?  0.49.18
What operating system are you using?  Windows Java 11

### Description
Load a Aerospace fighter, and head to the equipment tab. NPE hits.

### Files
```
13:59:31,936 ERROR [megameklab.MegaMekLab] {AWT-EventQueue-0}
megameklab.MegaMekLab.lambda$main$0(MegaMekLab.java:40) - Uncaught Exception Detected
java.lang.NullPointerException: Cannot invoke "megamek.common.Mounted.getLinked()" because "weapon" is null
	at megamek.common.WeaponType.getMaxRange(WeaponType.java:611)
	at megameklab.ui.util.EquipmentTableModel.getDamageString(EquipmentTableModel.java:471)
	at megameklab.ui.util.EquipmentTableModel.getValueAt(EquipmentTableModel.java:259)
	at java.desktop/javax.swing.JTable.getValueAt(JTable.java:2763)
	at java.desktop/javax.swing.JTable.prepareRenderer(JTable.java:5795)
	at java.desktop/javax.swing.plaf.basic.BasicTableUI.paintCell(BasicTableUI.java:2210)
	at java.desktop/javax.swing.plaf.basic.BasicTableUI.paintCells(BasicTableUI.java:2112)
	at java.desktop/javax.swing.plaf.basic.BasicTableUI.paint(BasicTableUI.java:1908)
	at java.desktop/javax.swing.plaf.ComponentUI.update(ComponentUI.java:161)
	at java.desktop/javax.swing.JComponent.paintComponent(JComponent.java:842)
	at java.desktop/javax.swing.JComponent.paint(JComponent.java:1119)
	at java.desktop/javax.swing.JComponent.paintChildren(JComponent.java:952)
	at java.desktop/javax.swing.JComponent.paint(JComponent.java:1128)
	at java.desktop/javax.swing.JViewport.paint(JViewport.java:736)
	at java.desktop/javax.swing.JComponent.paintChildren(JComponent.java:952)
	at java.desktop/javax.swing.JComponent.paint(JComponent.java:1128)
	at java.desktop/javax.swing.JComponent.paintChildren(JComponent.java:952)
	at java.desktop/javax.swing.JComponent.paint(JComponent.java:1128)
[...]

PR #1433, opened 2024-02-19
Fix 1429: mounted.getLinked() is null

Sleet01:
This is the MML addendum to the MM fix to #1429.
Technically it isn't necessary as the offending calls will be handled correctly once Megamek/megamek#5176 goes in, but this PR cleans the original calls up as passing `null` as a weapon value is ugh.

This PR also adds unit tests to:
- verify that every column of the `EquipmentTableModel` of each unit type can be populated and read;
- verify that `EquipmentTableModel.getDamageString()` (used to populate the damage field in the table) works correctly with Aerospace, CI, and Artillery;

said unit tests to be run via the "Verification/test" gradle task prior to any PR submissions.

There are still many, many areas within MML that need unit test coverage, but at least this gets the ball rolling.

NOTE:
The tests in this PR will continue to fail until Megamek/megamek#5176 is pulled.


SJuliez:
This PR contains a range of units. Is this intentional?

Sleet01:
> This PR contains a range of units. Is this intentional?

Yes, since the root issue only impacted the Aero equipment view, I wanted to add coverage for the full range of unit types in the new test to catch any similar single-type equipment view issues.

SJuliez (approved):

PR #1436, opened 2024-02-21
Fix integer overflow in determining maximum number of bay doors

### LINKED ISSUE #1242: Uncaught java exception? Attached all logs and unit file as the program requested.
Basically the title. Unsure what went wrong, or why, or if this is even a useful issue report, just... following directions. FWIW, the program went back to running after I saved the unit.
[logs_unitfile.zip](https://github.com/MegaMek/megameklab/files/11165788/logs_unitfile.zip)

neoancient:
When assigning the maximum value to the spinner for number of bay doors it adds the current number of doors on the bay to the number available for assignment. This works fine for aerospace units, which have maximum number based on weight, but support vehicles have no such maximum. The number of available doors is Integer.MAX_VALUE (I'm not sure what the -1 was for, but it prevented the issue from popping up as long as there was only one door). When added to the number of doors currently, it can result in integer overflow so that the maximum is a negative number. This causes an IllegalArgumentException because it doesn't meet the requirement that min <= value <= max.

Fixes #1242

### DIFF
--- a/megameklab/src/megameklab/ui/generalUnit/TransportTab.java
+++ b/megameklab/src/megameklab/ui/generalUnit/TransportTab.java
@@ -337,7 +337,7 @@ public class TransportTab extends IView implements ActionListener, ChangeListene
             }
             return Math.max(total, 0);
         } else {
-            return Integer.MAX_VALUE - 1;
+            return Integer.MAX_VALUE;
         }
     }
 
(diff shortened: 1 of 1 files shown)


SJuliez (approved):

Sleet01 (approved):
LGTM

PR #1448, opened 2024-03-05
Fix infantry weapon resetting on year change

### LINKED ISSUE #1432: Selecting year during infantry creation defaults primary weapon to auto-rifle.
0.49.18 (but was present in 0.49.17 as well).

While building infantry units, if the primary weapon is assigned to be anything other than an auto-rifle, selecting the year field and then clicking away (even if you don't change the year) causes the primary weapon to default to an auto-rifle.  Doesn't seem to impact support weapon selection (if any).

pavelbraginskiy:
Fixes #1432 

If the weapon became illegal with the year change, it actually *wouldn't* be reset.


mmetke-qumulo (approved):
Looks good.

Sleet01 (approved):
Looks good.

PR #1457, opened 2024-03-15
Print queue interface improvements

pavelbraginskiy:
After adding page breaks in a previous patch, I realized I found myself adding all my units first, then not being able to add page breaks where I want them.

Screenshot: 
image


pavelbraginskiy:
I'd like to replace the text of the buttons with nice up/down arrow icons, but I'm not sure how to accomplish such a thing.
I think it'd look nicer, if it was possible.

Sleet01:
Could add a gridbag to the left of the unit list, with graphical buttons (using icons) instead of words or Unicode characters.
Then you wouldn't need to do any Localization testing.

There should be some code in MML you could purloin to read and set icons; then it's just a matter of creating a thin grid of 4 x 1 buttons next to the "Selected Units" element.

pavelbraginskiy:
Added some icons I made myself.
New screenshot: 
image

Sleet01:
> Added some icons I made myself

Those.  Are.  _Awesome_.

SJuliez (approved):

PR #1462, opened 2024-03-18
When printing from MUL, calculate BV from C3/Tag

pavelbraginskiy:
Further builds on top of #1460 and additionally depends on MegaMek/megamek#5259.
To view the code changes just in this feature, select the last commit in this PR.
If #1460 is rejected, this PR could be refactored to be independent, but it does require the MegaMek part.

This PR makes it so that if printing from a MUL file, if the option to include pilot data is enabled, BV will be calculated with modifications from other units in the file, which includes C3 and TAG/semi-guided ammunition.

Testing:

A MUL file generated from the following force:
image

Produced the following prinout:
[Fury_Command_Tank__C3M__etc-job_56.pdf](https://github.com/MegaMek/megameklab/files/14630393/Fury_Command_Tank__C3M__etc-job_56.pdf)


pavelbraginskiy:
Checks fail because of missing MegaMek part.

PR #1464, opened 2024-03-19
Support Vees can mount sponsons

pavelbraginskiy:
MML previously wouldn't show Sponson turrets in the equipment database for support vehicles.

TO:AUE p160


Sleet01 (approved):
LGTM!
I wonder when this got changed, because I've definitely made support vehicles with sponson turrets...

pavelbraginskiy:
Turns out, support vehicles get sponson turrets from a checkbox in the Turrets section of the Structure tab, while Combat Vehicles get them from the equipment database. I think it's pretty silly that CVs and SVs have an entirely different UI for sponson turrets, but it does mean there was always a way for SVs to have sponsons.

PR #1478, opened 2024-04-10
Show a warning when attempting to print the invalid current unit

pavelbraginskiy:
I, and others I've heard from, occasionally make the mistake of creating a unit, forgetting to allocate some equipment or other, and then printing it, only to wonder where the equipment went. 

This change makes it so when printing the current unit, a warning is shown if it is invalid, same as if you tried to save it, because not everyone saves their units before printing them.

### DIFF
--- a/megameklab/src/megameklab/ui/MenuBar.java
+++ b/megameklab/src/megameklab/ui/MenuBar.java
@@ -362,7 +362,10 @@ public class MenuBar extends JMenuBar implements ClipboardOwner {
         final JMenuItem miExportCurrentUnitToPDF = new JMenuItem(resources.getString("CurrentUnit.text"));
         miExportCurrentUnitToPDF.setName("miExportCurrentUnitToPDF");
         miExportCurrentUnitToPDF.setMnemonic(KeyEvent.VK_U);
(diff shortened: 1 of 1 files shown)


megameklab/src/megameklab/ui/MenuBar.java
```diff
@@ -410,7 +413,10 @@ private JMenu createHTMLUnitExportMenu() {
         final JMenuItem miExportCurrentUnitToHTML = new JMenuItem(resources.getString("CurrentUnit.text"));
         miExportCurrentUnitToHTML.setName("miExportCurrentUnitToHTML");
         miExportCurrentUnitToHTML.setMnemonic(KeyEvent.VK_U);
-        miExportCurrentUnitToHTML.addActionListener(evt -> exportSummary(true));
+        miExportCurrentUnitToHTML.addActionListener(evt -> {
+            warnOnInvalid();
+            exportSummary(true);
```
SJuliez, line 418: Is it warning twice here with the warning in the called method?

pavelbraginskiy: Sure is, didn't even notice earlier. I'll get that corrected.

SJuliez (approved):

PR #1484, opened 2024-04-15
Fix #5252: critical assignment tab eq list missing keywords like "Ammo" and "[Half]" or "[Full]"

Sleet01:
Don't use the `.getShortName()` value for AmmoType instances in the Equipment List of the Crits tab.

Testing:
- Added offending ammo types from the bug to new units.
- Ran all three projects' unit tests.

Close MegaMek/megamek#5252

### DIFF
--- a/megameklab/src/megameklab/util/UnitUtil.java
+++ b/megameklab/src/megameklab/util/UnitUtil.java
@@ -1143,7 +1143,8 @@ public class UnitUtil {
      */
     public static String getCritName(Entity unit, EquipmentType eq) {
         String name = eq.getName();
-        if (name.length() > 22) {
+        // Only shorten non-ammo; getShortName leaves off "Ammo" and "[Half]" that we want
+        if (name.length() > 22 && !(eq instanceof AmmoType) ) {
             name = eq.getShortName();
         }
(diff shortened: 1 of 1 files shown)


HammerGS:
We’ll need to test with record sheets. As short name is used to make sure they print without some of the crazy l long names. 

I can do that latter today.

Sleet01:
@HammerGS Here's a JPG of a PostScript file saved from the MML print dialog:

pdf_output

The update doesn't seem to impact how the printout displays Ammo on sheets, just whether Ammo equipment in the Critical equipment list is shortened or not.

Sleet01:
@HammerGS Worst case scenario (200 ton SH Tripod loaded with as many ammos as possible):

Screenshot_20240415_122738
(All the ammo using regular names in the Crit Tab's equipment list)

pdf_output2
(Record Sheet is mostly legible, if a bit crowded)

HammerGS:
That's one intense unit. But that should work, mutators are always nasty for names and CGL doesn't use special ammos. So I'm good with it.

Thom293:
Does this solve https://github.com/MegaMek/megamek/issues/5159 too?  just saw it when looking for something else.

PR #1492, opened 2024-04-24
Trigger savefile selection when unit name changes

### LINKED ISSUE #1490: [RFE] Make <ctrl>-<s> open "Save As..." dialog when model or variant name is changed
### Environment
Version: 0.49.19
OS: Ubuntu 22.04 Linux
Java Version: 17.0.10

### Description
I've experienced this issue myself, and other users have related similar complaints:  When editing a unit in MML with the aim of creating a new variant file, after editing the model or variant name, the \<ctrl\>-\<s\> shortcut does not do the expected action of creating a new file using the new model and/or variant name, but instead overwrites the originally loaded file with the updated version.

This would be acceptable if the user were to leave the model and/or variant name as-is, indicating an update to the existing unit.  But changing either of those fields indicates that the user intends to create a new unit.

The requested functionality would be:

1. When \<ctrl\>-\<s\> (or \<alt\>-\<f\>, \<s\>; or File -\> Save) is selected, compose a new filename from the model and variant fields (as is done for Save As).
2. Choose whether to Save or Save As based on the generated filename:
2.1: If the newly-generated filename matches the current filename (the two name fields have not been changed) save over the original file.
2.2: If the newly-generated filename _does not_ match the current filename, fire the Save As event instead.

This has two benefits over our current save system:
A. users can avoid accidentally overwriting files when they mean to create new variants.
B. creating a new series of Omnimech configurations from a "base" or "template" config is simplified, as each configuration will differ in variant name from the base version.

This may seem like an inconsequential change from the current functionality, but will save quite a bit of work for users who need to do a lot of file editing and end up relying too much on muscle memory for loading and saving files.

pavelbraginskiy:
Fixes #1490

### DIFF
--- a/megameklab/src/megameklab/ui/MegaMekLabMainUI.java
+++ b/megameklab/src/megameklab/ui/MegaMekLabMainUI.java
@@ -33,6 +33,7 @@ public abstract class MegaMekLabMainUI extends JFrame implements RefreshListener
     private String fileName = "";
     protected MenuBar mmlMenuBar;
     protected boolean refreshRequired = false;
+    private String originalName = "";
     
     public MegaMekLabMainUI() {
         setDefaultCloseOperation(JFrame.DO_NOTHING_ON_CLOSE);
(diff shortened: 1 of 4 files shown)


Sleet01 (approved):
Looks good to me!

PR #1510, opened 2024-05-22
Improve readability of tables

pavelbraginskiy:
Adds shading to every other row of many of the harder to read tables to make them easier to scan horizontally.
In order to not affect official record sheets, shading must be enabled from an option in the Configuration menu.

An example of the shaded tables: 
image
See here for an example of every covered unit type: [testsheets.pdf](https://github.com/MegaMek/megameklab/files/15407987/testsheets.pdf)

### Problems
Enabling the option slows export/print of record sheets significantly, from 1-2 seconds to around half a minute per sheet. Since this doesn't affect users not using the option, which is the default, I don't think this is a blocker. I have no idea why this happens, and could use some insight.

The additional tables from *Print reference tables* and the fluff-image-replacement tables aren't shaded, and it would be nice if they were. These tables aren't in SVGs and would need a different, programmatic approach in order to be shaded.


Sleet01:
@pavelbraginskiy do you get the same slowdown if you select "Print" but then print to file?  This should produce a PostScript file (.ps) rather than a PDF, and may use a different code path for conversion.

pavelbraginskiy:
> @pavelbraginskiy do you get the same slowdown if you select "Print" but then print to file? This should produce a PostScript file (.ps) rather than a PDF, and may use a different code path for conversion.

I hadn't thought to try it, I assumed it would be the same way. *Exporting* is actually super fast, *printing* takes forever.

Still no idea why though...

Sleet01:
I copied one of your updated tables and (templates_us/tables_tank.svg) into 0.49.19.1 and didn't see any slowdown while exporting the file.  You may have an issue in your new code rather than the .svg conversion itself.

Sleet01:
> > @pavelbraginskiy do you get the same slowdown if you select "Print" but then print to file? This should produce a PostScript file (.ps) rather than a PDF, and may use a different code path for conversion.
> 
> I hadn't thought to try it, I assumed it would be the same way. _Exporting_ is actually super fast, _printing_ takes forever.
> 
> Still no idea why though...

This may be related to how PDFs represent filled rectangles, and the printer drivers.  I found several instances of people asking about this specific SVG use case and slowdowns.

pavelbraginskiy:
I wonder if it makes sense to rasterize the rectangles, then. Seems silly, but if it helps performance...

Sleet01:
> I wonder if it makes sense to rasterize the rectangles, then. Seems silly, but if it helps performance... I might experiment with it later.

It'd be good to get a sampling of other folks' print performance before doing much more work.  I'll do a proper review after lunch, then see if we can just pull this.  If it's broken for most tests we can consider more changes or fixes at that point.

Sleet01 (approved):
Looks good to me.
Recommend pulling and testing on as many different machines as possible to narrow down scope of printing/exporting slow-down.
Worst case, reverting just the .svg files will fix the issue.

pavelbraginskiy:
So I tried rasterizing the shading. It does improve performance back to near-instant, but it make the rendering of the text under the shading look _awful_. It isn't a solution.

Sleet01:
> So I tried rasterizing the shading. It does improve performance back to near-instant, but it make the rendering of the text under the shading look _awful_. It isn't a solution.

I blame HP; I don't have any specific reason, I am just willing to blame them :)

Let's see how the initial update works for folks, and then assess from there.

PR #1514, opened 2024-05-24
Fix attrocious table shading render speed

pavelbraginskiy:
Refactors the SVGs to avoid transparency, which apparently takes forever to print, probably because GDI has no transparency support.

Also improves the quality of the text when zooming all the way in, to boot!


Sleet01 (approved):
Looks great!

PR #1517, opened 2024-05-31
Add discord-formatted export support

### LINKED ISSUE #1513: [RFE] Export for Discord Option
This has been discussed a number of times, but I dont see an RFE so I figured I would memorilize it.

Simply, a "Export for Discord" option.  Basically just needs to add ``` at the start and end of the export, and perhaps auto-truncate fluff and/or spaces to fit under the free discord character cap.  Excluding fluff is usually enough, but sometimes on mechs with gobs of weapons, might need to truncate spaces too. 

If you want to be really fancy could include one of the markdown codes by default to pretty it up, but that is a pure gluttonous ask.  :)

image

Thank you.

pavelbraginskiy:
Closes #1513 

image

Currently just adapts the HTML formatting to Discord by applying bold and underline as appropriate, colors are technically supported but they're only used by the HTML export when the unit has damage.

Suggestions for sprinkling in additional colors welcome in a separate issue. 

Depends on MegaMek/megamek#5533

### DIFF
--- a/megameklab/resources/megameklab/resources/Menu.properties
+++ b/megameklab/resources/megameklab/resources/Menu.properties
@@ -46,6 +46,8 @@ htmlUnitExportMenu.text=To HTML
 textUnitExportMenu.text=To Text
 # Clipboard Unit Export Menu
 clipboardUnitExportMenu.text=To Clipboard
+# Discord Clipboard Unit Export Menu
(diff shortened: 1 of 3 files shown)


SJuliez (approved):

Thom293:
Just finally used this and it looks great.  Thank you.

PR #1520, opened 2024-06-10
Don't print Edge abilities of pilots

pavelbraginskiy:
Now that edge abilities are all on by default, printing record sheets from a MUL looks like this: 
image


IllianiCBT (approved):

IllianiCBT:
LGTM

PR #1521, opened 2024-06-11
Fix to-hit modifier for pilot damage on aerospace record sheets.

neoancient:
The optional reference tables on aerospace record sheets shows the hit modifier for pilot/crew damage as +2, but it should be +1/hit.


IllianiCBT (approved):
LGTM

neoancient:
I found two more issues with the aerospace reference tables:
1. Life support damage shows as +1/hit to control rolls, but per TW, p. 240, the modifier is +2 and any subsequent hits are ignored. This is also what the record sheets show.
2. Avionics hits are +1/hit for the first two hits, but on the third hit the avionics system is destroyed and the modifier becomes +5. I've added a line to show that, the same as with the sensors entries in the to-hit modifiers table.

IllianiCBT (approved):

PR #1540, opened 2024-06-30
Clean up of Actions and normalization of them.

rjhancock:
Removed actions not needed and fixed issues in nightly calls. Removed Maven Actions.


HammerGS:
Based on the Actions and the MM being approved merging this

PR #1560, opened 2024-07-14
Row shading for procedurally generated reference tables

pavelbraginskiy:
Previously row shading only existed for tables baked into the svgs of record sheet tables. 

This pr adds row shading for all the reference tables which are proceduraly generated. 

Care was taken to ensure the more unusually laid-out tables are shaded properly, but it's possible I missed something and will likely ask the QA testers to generate record sheets for all sorts of units and make sure they look ok with shading.

Shaded rows look like this:
image

The example is for vehicles, but all unit types are supported.


IllianiCBT (approved):
LGTM

PR #1568, opened 2024-07-22
Don't open RS right click menu on left click

pavelbraginskiy:


IllianiCBT:
While it is kinda obvious what this does, looking at the code, please make sure to include a description so users following on Discord can easily spot what each PR does. :)

IllianiCBT (approved):

PR #1573, opened 2024-07-25
Fix clipping issues for small unit RS preview

pavelbraginskiy:
The small unit sheet (Infantry, Battle Armor, and Protomechs) doesn't cope well with a nonstandard paper size. It gets cut off on the right side, and sometimes tables overlap with the unit blocks. 

This change uses standard margins for these unit types. This leads to an annoyingly large margin around the sheet, but it's better than straight up rendering wrong.


pavelbraginskiy:
Doesn't fix the bug completely, actual fix coming in a different PR

PR #1574, opened 2024-07-26
Record sheet fixes and improvements

pavelbraginskiy:
This PR collects several bug fixes and improvements that are somewhat related which I worked on at the same time.

- Row shading for tables looks better, the shading is now centered within the table.
- Record sheet preview for small units (CI/BA/Protos) was fixed to not cut off the right edge of the sheet.
- The option to replace mek fluff art with tables has been reworded to indicate it works on more than just meks.
- Small units will now include a cluster table instead of empty space when there aren't enough units to fill a whole sheet.
- The code that detects when a small unit sheet is full has been factored out into its own method and given unit tests.

### DIFF
--- a/megameklab/src/megameklab/printing/PrintSmallUnitSheet.java
+++ b/megameklab/src/megameklab/printing/PrintSmallUnitSheet.java
@@ -230,7 +230,7 @@ public class PrintSmallUnitSheet extends PrintRecordSheet {
         if (numTypes > 1) {
             throw new IllegalArgumentException("Heterogeneous unit types are not supported");
         }
-        if (entities.get(0) instanceof BattleArmor || entities.get(0) instanceof Protomech) {
+        if ((entities.get(0) instanceof BattleArmor) || (entities.get(0) instanceof Protomech)) {
             return entities.size() > 4;
         }
         if (entities.get(0) instanceof Infantry) {
--- a/megameklab/src/megameklab/printing/PrintSmallUnitSheet.java
+++ b/megameklab/src/megameklab/printing/PrintSmallUnitSheet.java
@@ -16,12 +16,17 @@ package megameklab.printing;
 import megamek.client.ui.swing.util.FluffImageHelper;
 import megamek.common.*;
 import megameklab.printing.reference.*;
+import org.apache.batik.anim.dom.SVGLocatableSupport;
+import org.apache.batik.bridge.BridgeContext;
+import org.apache.batik.bridge.DocumentLoader;
+import org.apache.batik.bridge.GVTBuilder;
+import org.apache.batik.bridge.UserAgentAdapter;
 import org.w3c.dom.Element;
 import org.w3c.dom.svg.SVGRectElement;
 
 import java.awt.*;
+import java.awt.geom.Rectangle2D;
 import java.awt.print.PageFormat;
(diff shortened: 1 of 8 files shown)


megameklab/src/megameklab/printing/PrintSmallUnitSheet.java
```diff
@@ -183,4 +191,51 @@ private void printBottomTable(ReferenceTableBase table, PageFormat pageFormat) {
...
+     */
+    public static boolean fillsSheet(List<? extends Entity> entities, RecordSheetOptions options) {
+        var numTypes = entities.stream().map(Entity::getClass).distinct().count();
+        if (numTypes == 0) {
+            return false;
+        }
+        if (numTypes > 1) {
+            throw new IllegalArgumentException("Heterogeneous unit types are not supported");
+        }
+        if (entities.get(0) instanceof BattleArmor || entities.get(0) instanceof Protomech) {
```
IllianiCBT, line 233: According to the style guide individual clauses in a conditional should be in separate parenthesis. i.e. `if `((cheese == 1) && (sausage == 2))` instead of `(cheese == 1 && sausage == 2)`

megameklab/src/megameklab/util/UnitPrintManager.java
```diff
@@ -155,23 +155,23 @@ public static List<PrintRecordSheet> createSheets(List<? extends BTObject> entit
                     }
                 } else if (unit instanceof BattleArmor) {
                     baList.add((BattleArmor) unit);
-                    if (singlePrint || baList.size() > 4) {
+                    if (singlePrint || PrintSmallUnitSheet.fillsSheet(baList, options)) {
```
IllianiCBT, line 158: Here too

megameklab/src/megameklab/util/UnitPrintManager.java
```diff
@@ -155,23 +155,23 @@ public static List<PrintRecordSheet> createSheets(List<? extends BTObject> entit
...
                         PrintRecordSheet prs = new PrintSmallUnitSheet(baList, pageCount, options);
                         pageCount += prs.getPageCount();
                         sheets.add(prs);
                         baList = new ArrayList<>();
                     }
                 } else if (unit instanceof Infantry) {
                     infList.add((Infantry) unit);
-                    if (singlePrint || infList.size() > (options.showReferenceCharts() ? 2 : 3)) {
+                    if (singlePrint || PrintSmallUnitSheet.fillsSheet(infList, options)) {
```
IllianiCBT, line 166: And here. Though you probably get the point.

megameklab/src/megameklab/util/UnitPrintManager.java
pavelbraginskiy: I don't think it makes sense to add parentheses here. There's only one operator in the new expression, although the code that was here before could have used them.

megameklab/src/megameklab/util/UnitPrintManager.java
pavelbraginskiy: Same here. `a || b` is fine, it doesn't need to be `(a) || (b)` unless `a` and `b` themselves have some operator in them.

IllianiCBT (approved):

PR #1582, opened 2024-08-01
Add toggle for RS adjusted BV printing, and set default export filename

### LINKED ISSUE #1581: RFE: Transfer MUL Name to PDF File Name
Currently working on a very large Record Sheet project. I have around 40 MUL files I'm printing from. 

Currently when you go to save the MUL you get this. 

image

But it would be a huge help to have this name become the default for the PDF
image

pavelbraginskiy:
Closes #1581.

This PR contains two features:
 - When exporting from MUL to PDF, prefill the name of the pdf with the name of the mul file.
 - Add a checkbox to the print queue dialog when printing from MUL to adjust the BV displayed on the record sheet for force modifiers like TAG and C3. BV is _always_ adjusted for pilot skill regardless of this checkbox.


HammerGS (approved):
Tested and working

PR #1595, opened 2024-08-21
Allow saving a MUL file from the queue-to-print dialog. 

### LINKED ISSUE #1592: Save MUL file from inside MML
I often get MUL files sent to me to print. But the MUL doesn't have the units in the correct order. So I need to rebuild the MUL to get the sheets in order. 

It would be nice from this screen to be able to resave the MUL. 

image

pavelbraginskiy:
Closes #1592.

In addition to letting you reorder the units in a MUL file, this also allows you to make a MUL file directly in MML, without involving MM, although you'll still need to involve MM to set pilot data or wire C3. 

This PR also enables you to add units from cache when printing from a MUL file to add additional units to it, but the additional units won't have pilot data. 

Eventually this dialog should have a way to customize units directly, but that's a more long-term goal.

### DIFF
--- a/megameklab/src/megameklab/ui/dialog/PrintQueueDialog.java
+++ b/megameklab/src/megameklab/ui/dialog/PrintQueueDialog.java
@@ -18,7 +18,9 @@
  */
 package megameklab.ui.dialog;
 
+import megamek.client.Client;
 import megamek.client.ui.baseComponents.MMButton;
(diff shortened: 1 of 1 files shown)


Sleet01 (approved):
LGTM.  Certainly I prefer re-using existing code in a simple and clear fashion over refactoring all the GUI stuff!

HammerGS:
Thank you for this!

PR #1597, opened 2024-08-22
Refactor saveUnitList() to not make a new ClientGUI

pavelbraginskiy:
Simple refactor of a previous pull request to not make the print-queue dialog depend on ClientGUI. 

Also sets the default filename of the new file to the name of the original file.

### DIFF
--- a/megameklab/src/megameklab/ui/dialog/PrintQueueDialog.java
+++ b/megameklab/src/megameklab/ui/dialog/PrintQueueDialog.java
@@ -19,6 +19,7 @@
 package megameklab.ui.dialog;
 
 import static java.util.stream.Collectors.toList;
+import static megamek.client.ui.swing.ClientGUI.CG_FILEPATHMUL;
 
 import java.awt.Container;
 import java.awt.Dimension;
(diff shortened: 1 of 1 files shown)


NickAragua (approved):

HammerGS:
Conflicting files.

megameklab/src/megameklab/ui/dialog/PrintQueueDialog.java
```diff
@@ -265,10 +270,34 @@ protected void okButtonActionPerformed(ActionEvent evt) {
...
+
+        File file = fileChooser.getSelectedFile();
+        if (!FilenameUtils.getExtension(file.getName()).equalsIgnoreCase(CG_FILEPATHMUL)) {
+            file = new File(file + "." + CG_FILEPATHMUL);
+        }
+
+        try {
+            EntityListFile.saveTo(file, entities);
+        } catch (IOException e) {
+            JOptionPane.showMessageDialog(getFrame(), "Error: " + e.getMessage());
```
rjhancock, line 299: Would suggest using the MMLogger with (exception, Message, title) if this is going to be kept.

Also suggest using String.format() for the message.

rjhancock: Honestly would suggest String.format() on any string concatenation.

pavelbraginskiy: Done

PR #1598, opened 2024-08-24
Add experimental option for drawing armor pips in groups of 5

pavelbraginskiy:
Many people prefer the armor pip layout used by programs like Flechs, which renders armor pips in clusters of 5. 

This pull request adds an option to attempt to render the armor pips in a similar fashion. 

Even with the option enabled, not all unit types are supported, units where the alternate pip-placement method is known to not work well will always use the old drawing method.

Sometimes adding pips in groups of 5 fails, in which case the code simply falls back to the old method.

The option to enable this method is clearly marked as Experimental so people don't think this feature is guaranteed to produce good results. A future version of MML will remove the Experimental marker and have better support for more units. 

I've playtested these sheets on the tabletop with players at my local LGS where they've been generally very well-received. 

The following sheets have dedicated support, with custom pip layout hints:

- WiGEs (0/1 turrets)
- Ground Vehicles (0/1/2 turrets)
- Aerospace Fighters
- VTOLs (0/1 turrets)
- Bipedal Mechs
- Tripod Mechs
- Quad Mechs
- Quadvees
- Aerodyne Dropships
- Spheroid Dropships
- Conventional Fighters
- LAMs
- Spheroid Small Craft
- Aerodyne Small Craft
- Naval vessels (0/1 turrets)
- Submarines (0/1 turrets)

The following sheets might be supported in the future, but are extremely rare and work _ok_ without dedicated support:

- All Superheavy Non-VTOL Vehicles
- Dualturret naval and submarine

The following units aren't supported, and armor pips will not be attempted to be drawn using the new method. These units mostly just have a completely different conception of how armor pips work for them, but for ProtoMeks specifically, it's because applying this method to their armor diagram causes the code to fail spectacularly in a way I haven't been able to really fix.

- Battle Armor
- Conventional Infantry
- Jumpships
- ProtoMechs
- Space Stations
- Warships

And now, some example record sheets:

image

image

image

image

image

image

image


pavelbraginskiy:
Thanks to TimberwolfD1 for providing SVG work to better support this armor layout scheme.

pavelbraginskiy:
Progress tracking notepad for unit support: https://hackmd.io/@oRoJxgEDQay55lwFkEPm3w/rk8lt4KiA/edit

rjhancock (approved):

rjhancock:
Just did a sanity check on code.

PR #1601, opened 2024-08-27
Fix 1588: NPE printing or previewing DropShip

### LINKED ISSUE #1588: Nightly: Dropship NPE - megamek.common.Aero.getEngine()" is null
### Environment
Linux
Java 17
MegaMekLab Nightly CI 1578

### Description
Received the NPE when trying to preview or export a PDF of the attached BLK file.

### Files
[RMS Titanic Titanic.blk.zip](https://github.com/user-attachments/files/16517461/RMS.Titanic.Titanic.blk.zip)
[megameklab.log](https://github.com/user-attachments/files/16517463/megameklab.log)

Sleet01:
Adds a safety check for the MML print code to prevent an NPE when an Aerospace unit's Engine isn't populated.
This _shouldn't_ happen, and is nominally controlled by MM code, but if it _does_, there's no real reason not to populate the sheet:

- DropShips don't even show the Engine info line,
- ASFs do, but it's informational, not functional

Testing:
- Ran checks with provided DS .blk files
- Ran all 3 projects' unit tests
- Added a unit test to confirm proper functionality in the above case.

Close #1588

### DIFF
--- a/megameklab/src/megameklab/printing/PrintAero.java
+++ b/megameklab/src/megameklab/printing/PrintAero.java
@@ -26,10 +26,9 @@ import org.w3c.dom.svg.SVGRectElement;
 import java.awt.*;
 import java.awt.geom.Rectangle2D;
 import java.awt.print.PageFormat;
-import java.io.File;
 import java.text.DecimalFormat;
-import java.util.*;
 import java.util.List;
+import java.util.*;
 import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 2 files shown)


repligator:
Did a bit of testing this morning with the latest MM/MML. You can reliably reproduce the NPE from https://github.com/MegaMek/megameklab/issues/1588 by going to new Warship, changing the Vessel Type to Subcompact Warship, going to preview, and pressing Record Sheet.

When I tested under this PR, the NPE is no longer thrown when repeating the above steps.

PR #1620, opened 2024-09-21
Frameless sheets

pavelbraginskiy:
Depends on #1598. Doesn't use any features from that PR per se, but it touches the same files, and this prevents merge conflicts. If that PR hasn't been merged yet, [this commit](https://github.com/MegaMek/megameklab/commit/47bf3242586a2872cef69ac1aa3f405e7988a479) is the one with all the code changes for this PR.

Adds an option to draw record sheets without frames, as requested by CGL.
Support all the 'Mechs (Bipedal, Quad, Tripod, QuadVee, LAM, and SuperHeavy variants).

Since this feature isn't expected to be useful for our non-CGL users, the ISO paper and TO heat options disable frameless output entirely, and the row shading option produces bad output. 

image
image
image
image
image


HammerGS (approved):
Tested and seems to work

rjhancock:
If this has all of the changes in #1598 as well as CGL's request, Id' suggest closing the other and making this one be all of the changes.

rjhancock (approved):

PR #1625, opened 2024-09-27
Print SPAs in record sheets printed from MUL

pavelbraginskiy:
This seems to be a feature that previously existed in MML, but at some point got broken.

When including pilot data when printing from a MUL, print the unit's SPAs in the Warrior Data box.
image

This has no impact on official sheets, since official sheets don't include warrior data.


HammerGS (approved):
Tested and seems to work

rjhancock (approved):

PR #1628, opened 2024-09-29
Fix iso weapons inventory

### LINKED ISSUE #1325: [0.49.15] ISO A4 size PDF export doesn't have enough room for weapon ranges
This issue doesn't appear in the current stable release (0.48.0), however fonts aren't correctly used in that version (mix of selected fonts and Times).

### Environment
This issue is present in both Standalone and HQ MekLab. It doesn't appear to be OS-specific; I've confirmed the issue on both MacOS and Windows.

### Description
When exporting PDF in US Letter size:
image

When exporting PDF in ISO A4 size:
image

This doesn't appear to be font-specific; I have the same issue when using other fonts.

pavelbraginskiy:
Fixes #1325.
When the new display for the armor type was added, some elements of the ISO mech sheets got mismatched with the size of the box they're in. This fixes their sizes so ISO mech sheets look normal again.

See linked issue for Before image.
After:
image


HammerGS (approved):
Tested and seems to work.

Issue #1631, opened 2024-10-04
[0.50.0 and on] Support Vehicles Become Invalid When Saved, BAR Level Edition


WeaverThree:
### Prerequisites and Pre-Issue Checklist

- [X] I'm reporting the issue to the correct repository:
- [MegaMek](https://github.com/MegaMek/megamek/issues)
- [MegaMekLab](https://github.com/MegaMek/megameklab/issues)
- [MekHQ](https://github.com/MegaMek/mekhq/issues)

- [X] I've tested the issue against at least the latest MILESTONE version
- [X] I've asked on the MegaMek Discord about the error
- [X] I've reviewed the BattleTech rules and MegaMekLab documentation, and I've confirmed that something isn't working as intended.
- [X] I've searched the Github tracker and haven't found the issue listed

### Severity *

High (Major Disruption): A major feature is broken or incorrect, but a workaround exists.

### Brief Description *

When you save a support vehicle, its armor BAR and tech ratings change such that the armor weight is larger and causes the vehicle to be overweight, and the armor is too dense to be used without the armored chassis mod. Details below.

Tested in 0.50.0 and a recent nightly.

It's likely this is related to #1596.
It's possible this is related to #1489.
This might be what's causing the Sherpa variants to be invalid, or other support vehicles.

### Steps to Reproduce

1. Start MML and hit "New Support Vehicle"
2. First Page: Set Tonnage to 50, Cruise MP to 5, Fuel Mass to 5. Give the unit a Chassis and Model for the filename.
3. Armor page: Select BAR 7 Armor, maximize armor, auto-allocate armor. The armor tech level will be D by default.
4. Transport page: Click 'Use Remaining Tonnage for Cargo'
5. Press Validate Unit - validation passed
6. Save the file. Load the saved file. Say you don't want to save the file again.

The warning dialog will appear saying:
![2024-10-04_103051](https://github.com/user-attachments/assets/001e3ff3-c30b-4d39-8d96-9fbae74f11e9)

7. Hit OK and note that the armor tonnage on the first page has jumped up.
8. Go to the Armor Page again. Note that the armor type has changed to BAR 2, tech rating B, and the tonnage has increased.
9. You can now only select BAR 2 or BAR 3 armor.
10. Select 4 tons of rating B BAR 3 armor and hit Auto-Allocate
11. The unit is now under tonnage so go to the transport page and add another ton of cargo space.
12. Save and load the file again. It works this time, without being invalid.

These are the relevant blk files:

[...]

PR #1650, opened 2024-11-03
Fix 1631: Armor rating incorrectly persisted

Scoppio:
The unit encoding process looks for a random (hardcoded the number 1) "armor tech level" inside the array of the armor_tech_level and picks that value as the armor tech level, however this value isnt being set when the armor is changed, the armor_tech_rating is set instead.
I need to validate why or how the BAR is set, but it is being correctly set, so this isnt as much a big of a deal.

Also, it had a piece of code right after the part where it should be setting it, but it wasan't correct and also was never called by any code anywhere, so I removed it.

Maybe I should take a further look in the persisting process because it is kind of a messs right now.

Fixes: #1631

### DIFF
--- a/megameklab/src/megameklab/ui/supportVehicle/SVArmorTab.java
+++ b/megameklab/src/megameklab/ui/supportVehicle/SVArmorTab.java
@@ -127,17 +127,8 @@ public class SVArmorTab extends ITab implements ArmorAllocationListener {
     @Override
     public void armorTechRatingChanged(int techRating) {
         getEntity().setArmorTechRating(techRating);
+        getEntity().setArmorTechLevel(techRating);
         getEntity().recalculateTechAdvancement();
-        panArmor.setFromEntity(getEntity());
-        panArmorAllocation.setFromEntity(getEntity());
-        refresh.refreshSummary();
-        refresh.refreshStatus();
-        refresh.refreshPreview();
-    }
-
-    @Override
-    public void armorBARRatingChanged(int bar) {
-        getEntity().setBARRating(bar);
         panArmor.setFromEntity(getEntity());
         panArmorAllocation.setFromEntity(getEntity());
         refresh.refreshSummary();


pavelbraginskiy (changes requested):
This doesn't really fix the issue. The bug can occur even when armorTechRatingChanged is never called (see: https://github.com/MegaMek/megameklab/issues/1631#issuecomment-2397589827).
It's also wrong to setArmorTechLevel to a tech rating constant, the two are different. I think you got lucky in testing and happened to set the tech level to a value that makes broken code *elsewhere* behave right.

Scoppio:
When I arrive home I will check that, but it does seen to be doing what is expected, since this is what the TestSupportVehicle exact weight calculator, so I may have been mislead for taking it as source of truth.

pavelbraginskiy:
I think a good place to start looking would be to load a support vehicle and immediately encode it to BLK, before anything in MML can happen to it, and see if there's a difference. If there is, then the bug is in BLK encoding and decoding (although there might still be something in MML causing related secondary bugs).

Scoppio:
> I think a good place to start looking would be to load a support vehicle and immediately encode it to BLK, before anything in MML can happen to it, and see if there's a difference. If there is, then the bug is in BLK encoding and decoding (although there might still be something in MML causing related secondary bugs).

I did exactly that, but I missed one thing, thanks, the solution I came up with could be "correct" by coincidence, but I am unsure...

the unit is loaded and have the value for armorTechlevel to be [1,1,1,1,1,1,1] which is aparently set as the value for the armor_tech_level overall since this should not be any kind of special armor. This is basic Inner Sphere intro tech, it has to be 1. (
Bar7 says that its static tech level is 1, so this correct, so this should not be the problem).

```java
} else if (dataFile.exists("barrating")) {
    megamek.common.equipment.ArmorType armor = ArmorType.svArmor(dataFile.getDataAsInt("barrating")[0]);
    sv.setArmorType(armor.getArmorType());
    sv.setBARRating(armor.getBAR());
    sv.setArmorTechLevel(armor.getStaticTechLevel().getCompoundTechLevel(false));
} else {
```

here armorTechlevel sets to all ones, which is correct.

then when it runs `warnOnInvalid(loadedUnit)` called inside loadUnitFromFile, the validateUnit from UnitUtil gets the entity verifier and runs:
```java
testEntity.correctEntity(sb, unit.getTechLevel());
```
There we have the error happening in the first test line in the TestSupportVehicle in the "correctEntity" method.

It then calls `calculateWeightExact()` who calls `getWeightArmor()` who calls ```
    @Override
    public double getWeightArmor() {
        return supportVee.getArmorWeight();
    }
```

And here we have this:

```Java
 public double getArmorWeight() {
        if (hasPatchworkArmor()) {
            double total = 0;
            for (int loc = 0; loc < locations(); loc++) {
                total += getArmorWeight(loc);
            }
            return RoundWeight.standard(total, this);
        } else {
            ArmorType armor = ArmorType.forEntity(this);
            if (armor.hasFlag(MiscType.F_SUPPORT_VEE_BAR_ARMOR)) {
                double total = getTotalOArmor() * armor.getSVWeightPerPoint(getArmorTechRating());
                return RoundWeight.standard(total, this);
            } else if (armor.hasFlag(MiscType.F_BA_EQUIPMENT)) {
                return getTotalOArmor() * armor.getWeightPerPoint();
            } else {
                double armorPerTon = ArmorType.forEntity(this).getPointsPerTon(this);
                return RoundWeight.standard(getTotalOArmor() / armorPerTon, this);
            }
        }
    }
```

The part that interest us is just this one:
```java
ArmorType armor = ArmorType.forEntity(this);
if (armor.hasFlag(MiscType.F_SUPPORT_VEE_BAR_ARMOR)) {
    double total = getTotalOArmor() * armor.getSVWeightPerPoint(getArmorTechRating());
    return RoundWeight.standard(total, this);
}
```

The call for `getArmorTechRating()` is looking for the value in the property `armorTechRating`, which is persisted as `armor_tech`

the code that adds the property and value `armor_tech` is this part:
https://github.com/MegaMek/megamek/blob/f19b6591e62347036925e2aef488ca4e1c61ba01/megamek/src/megamek/common/loaders/BLKFile.java#L749

which looks fot the the armorTechLevel of the armor of the unit in a pre-specified location (hardcoded to be location 1), not the tech rating, but the tech level...

Currently, the value in this place is always 1, because it should be 1 since it is an Intro tech armor type
However, the number we are looking for in this case is 3, the rating (A, B, C, D, E, F), not the tech level (1).

If we load the rating 1 it will multiply the number of points of armor for the value of the wrong index:
```java
armor.weightPerPointSV = new double[] { 0.180, 0.088, 0.056, 0.045, 0.040, 0.037 };
```
which is the index 1, so 104 * 0.088 = 9.152, which rounds to 9.5 tons, and therefore breaks everything.

If it reads the rating 3, the value is 104*0.045, which is 4.69, rounds to 5 tons, the expected amount.

Thats why my initial assumption worked, the index match, however, there can only be one error here, either armor_tech_rating should be a value present and should be set at 3, or armor_tech should be present and should be at a value of 3.

From all that I could get, the armorTechLevel is correct, and so are you, I missed in my fix here and the result was a coincidence. But this might mean that we have a problem with the BLK Encoder on MegaMek, encoding the wrong value there?
Please let me know what you think of that one.

PR #1655, opened 2024-11-06
feat: adds a test that does load after save and save after load

Scoppio:


HammerGS:
This is a year and the dev is no longer with the projects.

PR #1663, opened 2024-12-10
feat: inverts one function check

Scoppio:
This function is called inverted everywhere it is called, so I just simplified it.

This is part of the acs-autoresolve

To be merged together with the other acs-autoresolve branches.


Sleet01 (approved):
LGTM

PR #1665, opened 2024-12-14
feat: inverted function call

Scoppio:
This Pull request is part of an overarching change on Megamek to extract an options interface.
This pull request depends on the PR of the same name on Megamek, which adds the inverted call for this function. 

### Related

https://github.com/MegaMek/megamek/pull/6265


Sleet01 (approved):
Changes look good, but it would be nice to have some info in the comment, or a link to a central PR if this is part of a larger set of work.

PR #1673, opened 2024-12-31
Allow opening multiple units in tabs

pavelbraginskiy:
Many people have requested the ability to have many units open at once.

image

Everything _seems_ to just work the way you might expect it to, but I'd like to get this to the testers to try to break it as soon as possible.

This shouldn't impact MekHQ, and when I opened MHQ MML it didn't instantly explode, but I don't know how to use MHQ so testing is necessary to make sure everything really does still work there.

You can shift-click the close-tab button to skip the "Would you like to save" dialog.

When you close MML, the state of the tabs is saved to a hidden directory in the user folder.

If the startup mode is set to "Restore Tabs", a user can pick up right where they left off, even including any unsaved work. 

Future work:

- Allow opening all the units from a MUL file in tabs. 
- Include a setting to periodically save your tab state, so you can recover your work even if your computer crashes.


megameklab/src/megameklab/ui/util/TabStateUtil.java
```diff
@@ -0,0 +1,172 @@
...
+        try (
+            var fos = new FileOutputStream(filenameAssociationsFile);
+            var ps = new PrintStream(fos)
+        ) {
+            for (var entry : filenameAssociations.entrySet()) {
+                ps.print(entry.getKey().getPath() + '\0' + entry.getValue() + '\0');
+            }
+        }
+
+        clean.createNewFile();
```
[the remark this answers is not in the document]
pavelbraginskiy: This is fine: The presence of the `clean` file indicates that the rest of the process concluded without issue. If creating this file somehow fails, we just don't try to restore tabs next time.

megameklab/src/megameklab/ui/util/TabStateUtil.java
```diff
@@ -0,0 +1,172 @@
...
+        }
+
+        var parts = Files.readString(Paths.get(db.getAbsolutePath())).split(Pattern.quote("\0"));
+        for (int i = 0; i < parts.length; i += 2) {
+            var entityFile = new File(parts[i]);
+
+            var newFile = new File(entityFile.getAbsolutePath().replaceFirst("\\.tmp$", ""));
+            FileUtils.copyFile(entityFile, newFile);
+
+            var fileName = parts[i + 1];
```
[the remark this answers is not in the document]
pavelbraginskiy: This can only happen with a malformed database, in which case crashing here would be the right thing to do.

Sleet01:
@pavelbraginskiy Can you grab the `rs_font` entry from your `megameklab.properties` file?  I don't seem to have any fonts that support the "➕" and "❌" symbols.

megameklab/src/megameklab/ui/MegaMekLabTabbedUI.java
```diff
@@ -0,0 +1,352 @@
...
+        setTransferHandler(new MMLFileDropTransferHandler(this));
+
+        // Remember the size and position of the window from last time MML was launched
+        pack();
+        restrictToScrenSize();
+        setLocationRelativeTo(null);
+        CConfig.getMainUiWindowSize(this).ifPresent(this::setSize);
+        CConfig.getMainUiWindowPosition(this).ifPresent(this::setLocation);
+
+        // ...and save that size nad position on exit
```
Sleet01, line 98: "and"

megameklab/src/megameklab/ui/MegaMekLabTabbedUI.java
```diff
@@ -0,0 +1,352 @@
...
+                TabStateUtil.saveTabState(editors.stream().limit(editors.size() - 1).toList());
+            } catch (IOException e) {
+                throw new RuntimeException(e);
+            }
+        }
+
+        return true;
+    }
+
+    private void restrictToScrenSize() {
```
Sleet01, line 253: "Screen"

megameklab/src/megameklab/ui/MegaMekLabTabbedUI.java
```diff
@@ -0,0 +1,352 @@
...
+    @Override
+    public void refreshMenuBar() {
+        menuBar.refreshMenuBar();
+    }
+
+    @Override
+    public MenuBar getMMLMenuBar() {
+        return menuBar;
+    }
+    //</editor-fold>
```
Sleet01, line 293: I don't think IDE-specific formatting comments should be included in the main repo.

megameklab/src/megameklab/ui/MegaMekLabTabbedUI.java
```diff
@@ -0,0 +1,352 @@
...
+     * Represents a button used for creating new tabs in the MegaMekLabTabbedUI interface.
+     * This class extends JPanel and is rendered as a non-opaque panel containing a "+" symbol.
+     * Used to mimic functionality for adding new tabs in a tabbed user interface.
+     * <p>
+     * The only reason this is a separate class instead of just the text for a tab is so that it can be green.
+     */
+    private static class NewTabButton extends JPanel {
+        public NewTabButton() {
+            setOpaque(false);
+            var label = new JLabel("➕");
```
Sleet01, line 306: Request that you add:
```
            label.setFont(new Font("Symbola", Font.PLAIN, label.getFont().getSize()));
 ```
so that non-Windows users can see the symbols.

File: megameklab/src/megameklab/ui/MegaMekLabTabbedUI.java [hunk 7]
```diff
@@ -0,0 +1,352 @@
...
+        MegaMekLabMainUI editor;
+
+        public ClosableTab(String name, MegaMekLabMainUI mainUI) {
+            unitName = new JLabel(name);
+            editor = mainUI;
+
+            setOpaque(false);
+
+            closeButton = new JButton("❌");
+            closeButton.setForeground(Color.RED);
```
Sleet01, line 330: Request that you add:
```
            closeButton.setFont(new Font("Symbola", Font.PLAIN, closeButton.getFont().getSize()));
 ```
so that non-Windows users can see the symbols.

File: megameklab/src/megameklab/ui/util/TabStateUtil.java [hunk 8]
```diff
@@ -0,0 +1,172 @@
...
+            File unitFile;
+            if (editor.getEntity() instanceof Mek) {
+                unitFile = File.createTempFile("mml_unit_", ".mtf.tmp", dir);
+                try (
+                    var fos = new FileOutputStream(unitFile);
+                    var ps = new PrintStream(fos)
+                    ) {
+                    ps.println(((Mek) editor.getEntity()).getMtf());
+                } catch (Exception e) {
+                    continue;
```
Sleet01, line 72: This definitely needs logging of the exception; no silent file errors.

megameklab/src/megameklab/ui/util/TabStateUtil.java
```diff
@@ -0,0 +1,172 @@
...
+                }
+            } else {
+                unitFile = File.createTempFile("mml_unit_", ".blk.tmp", dir);
+                try {
+                    BLKFile.encode(unitFile.getPath(), editor.getEntity());
+                } catch (EntitySavingException e) {
+                    continue;
```
Sleet01, line 79: Same as above.

megameklab/src/megameklab/ui/util/TabStateUtil.java
```diff
@@ -0,0 +1,172 @@
...
+            try {
+                Entity loadedUnit = new MekFileParser(newFile).getEntity();
+                newFile.delete();
+                var editor = UiLoader.getUI(UnitUtil.getEditorTypeForEntity(loadedUnit), loadedUnit.isPrimitive(), loadedUnit.isIndustrialMek());
+                editor.setEntity(loadedUnit);
+                editor.setFileName(fileName);
+                editor.reloadTabs();
+                editor.refreshAll();
+                editors.add(editor);
+            } catch (EntityLoadingException ignored) {}
```
Sleet01, line 139: We definitely want logging of exceptions while attempting to read unit files.

megameklab/src/megameklab/ui/util/TabStateUtil.java
```diff
@@ -0,0 +1,172 @@
...
+            FileUtils.copyFile(entityFile, newFile);
+
+            var fileName = parts[i + 1];
+            if (fileName.isBlank()) {
+                fileName = "";
+            }
+
+            try {
+                Entity loadedUnit = new MekFileParser(newFile).getEntity();
+                newFile.delete();
```
Sleet01, line 132: I'd argue that we shouldn't delete any files until we've completed all actions in the `try` block.

Sleet01 (changes requested):
In addition to the stuff I commented on, I think we need to add the standard tab-related menu items and symantics:
- File -> Close, to mirror the X button function;
- File -> Open, similar to load but replaces the current tab's contents with the selected unit.
- Wrap the selected tab around instead of creating infinite new tabs, when using CTRL+Tab and right arrow in the tab bar.

Otherwise, this looks fantastic and I'm hyped to start using the tabs!

Sleet01 (approved):
Thanks, looks really good!

PR #1676, opened 2025-01-02
Fix: Tank armor and invalid builds

Scoppio:
Implements test suite to validate all units in our current data folders.


HammerGS:
Given units are based in MegaMek for the builds etc, should these be a MegaMek Side fix?

PR #1684, opened 2025-01-06
Refresh menu bar to new current unit when new tab is selected

pavelbraginskiy:
You can't switch unit type to the unit type of the current unit, but if the menu bar isn't refreshed, it'll use the unit type of the _first_ unit, not the _current_ unit, to determine which unit kind you can't switch to.

Refreshing the menu bar when you switch to a new tab solves this (and likely some other problems as well).


Sleet01 (approved):
LGTM!

PR #1685, opened 2025-01-07
Detect unsaved work

pavelbraginskiy:
This enables MML to determine if a unit has unsaved changes, offering us two features: First, tabs now show a little * next to the unit name if the tab has unsaved work, and second, we don't show the "would you like to save your work" safety prompt for units which don't have any work to save.

This PR also contains a very small extra feature, which allows you to middle click tabs to close them, which might cause merge conflicts if it was in a separate PR.

### DIFF
--- a/megameklab/src/megameklab/ui/MegaMekLabTabbedUI.java
+++ b/megameklab/src/megameklab/ui/MegaMekLabTabbedUI.java
@@ -23,12 +23,13 @@ import megamek.MegaMek;
 import megamek.client.ui.swing.util.UIUtil;
 import megamek.common.Entity;
 import megamek.common.preference.PreferenceManager;
-import megameklab.EntityChangedUtil;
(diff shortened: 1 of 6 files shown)


megameklab/src/megameklab/ui/MegaMekLabTabbedUI.java
```diff
@@ -88,6 +91,11 @@ public MegaMekLabTabbedUI(MegaMekLabMainUI... entities) {
         // Enable opening unit and mul files by drag-and-drop
         setTransferHandler(new MMLFileDropTransferHandler(this));
 
+        // If you can think of a way to detect when the user makes a change to the entity, let me know.
```
Sleet01, line 94: We might be able to get by with a 'dirty' flag in the entity or something, but for now let's see how this works out.

Sleet01 (approved):

PR #1688, opened 2025-01-09
Fix font rendering of symbols on windows using `gradle run`

pavelbraginskiy:
Instead of just hoping the system has Symbola or a suitable fallback font available, this forces the use of Noto Sans Symbols 2, a font we already include with MegaMek. 

This fixes the issue that some icons wouldn't render properly when using `gradle run` on Windows.


Sleet01 (approved):
Looks good!  Thanks for taking point on this.

Sleet01:
@pavelbraginskiy looks like there's a conflict since merging the updated-tab PR.

Sleet01 (approved):
LGTM

PR #1694, opened 2025-01-15
Revert unsaved work detection

pavelbraginskiy:
There were a billion problems with the detection of unsaved work, and some of them appear unfixable without a _major_ refactor of many systems.


IllianiCBT:
This is a shame, but based on your description it sounds like the right move

PR #1697, opened 2025-01-18
Tabbed UI Polish Pass

pavelbraginskiy:
Lots of small fixes and QOL improvements:

- Ensure that renaming a unit renames its tab, even after switching unit type.
- Only show the "Are you sure" prompt for operations where it makes sense, i.e. not loading a unit in a new tab.
- Fix "Load from file" not working from the startup GUI.
- Allow you to close tabs by middle-clicking them.


Sleet01 (approved):
LGTM

PR #1709, opened 2025-02-03
Added Infantry `Mob` Unit Files

IllianiCBT:
- Added "Mob" unit files for Small, Medium, Large, and Huge variants. Created by @HammerGS.

These are used by the upcoming Prisoner rework.


rjhancock:
These only need to be added to the main MM repo. The build scripts copy them over from there.

IllianiCBT:
OH! Oh that's cool! I assume that means I can also close off the mhq version of this?

rjhancock:
Yes.

Every now and then @HammerGS will merge the data from MM into MML and MHQ. I'm still working to find a way to put data separate from the rest of the repos so we only have to update one place.

IllianiCBT:
Ok, neato. I’ll get this and the mhq copy closed off when I’m back from running chores

PR #1712, opened 2025-02-04
Allow locking the base chassis of an Omnimek

pavelbraginskiy:
To make it easier to build new configurations of existing OmniMeks, this makes it possible to prevent all modifications to the base chassis while allowing pod equipment to be altered.

This should prevent modifying any of the Mek's systems, removing or moving any fixed equipment, adding any equipment that can't be pod-mounted, or adding any equipment as fixed equipment.

For now, only OmniMeks are supported. OmniVehicles and OmniFighters will hopefully be covered in the future, but admittedly my desire to work on those is completely tapped at the moment. 

image


repligator:
Gladiator (Executioner) Prime (ID 1211)
I can remove the (fixed) heat sinks by pressing up or down on the Structure/Armor tab under heat sinks.

I'm also getting an NPE when right-clicking on limbs under the Assign Criticals tab.
`java.lang.NullPointerException: Cannot invoke "megamek.common.CriticalSlot.getMount()" because "cs" is null`
[megameklab.log](https://github.com/user-attachments/files/18673211/megameklab.log)

pavelbraginskiy:
The fact that the total number of heatsinks can go lower than the number of base heatsinks has always been an oddity, but I it'll be good to get that patched up anyway.

PR #1728, opened 2025-02-15
Improve RS display of information for Insulators/Pulse Modules/Capacitors

pavelbraginskiy:
Previously Laser Insulators and RISC Laser Pulse Modules would be listed as ordinary equipment in the RS inventory readout.
This adjusts them to be displayed together with the weapon they're associated with, similar to how Artemis and PPC Capacitors work.

Additionally, since these enhancements (plus the capacitor) affect the weapon's basic stats and they take up a second line on the sheet, we can use that second line to display the new stats for for the weapon with that enhancement. 

Example sheet:
image

In the future it would be nice to also update the value shown in the heat profile, but that's a separate (and more complex) issue that I won't be addressing for now.


SJuliez:
Code is all good. Looks like a good idea to me, but needs approval by Hammer.

PR #1739, opened 2025-02-21
Chain Drape

### LINKED ISSUE #1222: Chain drape
The Chain Drape is a new piece of mech equipment introduced in PseudoTech: Arcade Operations: UrbanFest (page 16) and should be implemented.

pavelbraginskiy:
Chain Drape from Arcade Operations: UrbanFest.

Closes #1222.

Requires MegaMek/megamek#6594, see there for more information.

There are a couple differences between the published UrbanKnight sheet and this recreated one.
First, the printed MP is reduced to be consistent with Modular Armor reducing the MP listed on the record sheet.
Second, the BV doesn't match. There is no published information on calculating BV for the chain drape, so I don't know how to make it line up.
image

### DIFF
--- a/megameklab/src/megameklab/util/MekUtil.java
+++ b/megameklab/src/megameklab/util/MekUtil.java
@@ -640,7 +640,7 @@ public final class MekUtil {
                     locations.add(Mek.LOC_CLEG);
                     blocks = 3;
                 }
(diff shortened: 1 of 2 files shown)


Sleet01 (approved):
LGTM!  Still needs to wait on MegaMek/megamek#6594, though.

PR #1742, opened 2025-02-25
Handheld weapons Part the First: Loading and Construction

pavelbraginskiy:
Requires megamek/MegaMek#6617.

image


megameklab/src/megameklab/ui/generalUnit/StatusBar.java
```diff
@@ -76,6 +76,10 @@ public StatusBar(MegaMekLabMainUI parent) {
         JButton btnValidate = new JButton("Validate Unit");
         btnValidate.addActionListener(validationListener);
 
+        JButton btnRefresh = new JButton("Refresh UI");
+        btnRefresh.setToolTipText("Refresh the UI, possibly repairing it if it is in a broken state.");
+        btnRefresh.addActionListener(evt -> refresh.refreshAll());
+
```
Sleet01, line 82: This seems to indicate something else isn't configured correctly.
What is causing the UI to need a refresh?

megameklab/src/megameklab/ui/handheldWeapon/HHWStructureTab.java
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    public void refresh() {
+        removeAllListeners();
+        panBasicInfo.setFromEntity(getEntity());
+        panChassisView.setFromEntity(getEntity());
+        panEquipmentView.refresh();
+        panSummary.refresh();
+        panIcon.refresh();
+        addALlListeners();
```
Sleet01, line 115: Should be `addAllListeners()`

megameklab/src/megameklab/ui/handheldWeapon/HHWStructureTab.java
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    public void removeAllListeners() {
+        panBasicInfo.removeListener(this);
+        panChassisView.removeListener(this);
+    }
+
+    public void addALlListeners() {
```
Sleet01, line 123: Typo.

Sleet01 (changes requested):
One tiny typo, and I'm curious about the UI refresh, but otherwise all looks really good.

megameklab/src/megameklab/ui/generalUnit/StatusBar.java
pavelbraginskiy: I added this button for myself to help debug problems (is the internal state messed up or did I miss a refresh?), but then I realized that there have been plenty of UI bugs in MML outside of handheld weapons that are caused by failing to refresh something, and as a user, I would have liked to have had this button on several occasions.

This is technically a feature completely unrelated to the HHW implementation, I just thought it would be helpful to have, because experience says that there were refresh bugs in MML before this PR and there will be new ones introduced in the future.

PR #1750, opened 2025-03-01
Rename Size/Shots to Shots for HHWs

pavelbraginskiy:
No resizable equipment can be mounted on a HHW, so it doesn't make sense to label the column Size/Shots when it can only ever be the number of shots for an ammo bin.

Update the default HHW sprite for MML while we're at it.


psikomonkie (approved):
Nice!

PR #1754, opened 2025-03-03
Print all open units + Performance Pass

### LINKED ISSUE #1747: [RFE] Queue all open units to print
### Brief Description of Enhancement *

It should be possible to queue the unit in every open tab for printing at once with a "Print All" option in the toolbar menu.

### Use Cases or Rationale

Following #1746, once units have been sent to the editor for review, if any changes have been made, it should be convenient to send them all back for printing.

pavelbraginskiy:
Closes #1747.

This PR has two major features. I would have created them as two separate PRs, but unfortunately they both touch tabbed UI internals so they need to be coordinated. See individual commits for each feature.

1. Adds a button to the menu bar under Print/Export Current Unit to Print/Export All Open Units. This collects every unit from all open tabs and sends them to the print-queue dialog, which already supported being prefilled with units in order to be used with MUL files.
2. Refactor and performance pass of MML unit loading. Two major gains were made here. First, we no longer create useless blank Mek tabs which are fated to instantly be replaced by an editor for the unit type we actually intended to load. Second, the equipment database was massively sped up by caching the DecimalFormat it used in sorting. The DecimalFormat constructor is incredibly slow, it turns out.

This PR also accomplished a couple other minor tasks that were too small for their own PR and made their way into the Print-all-units commit:
- When exporting units, the print-queue dialog now says Export instead of Print on buttons and the window title.
- Removed some dead code from the `EditorTab` class which was left over from a feature that has since been removed.


IllianiCBT:
I can see our in-person tournament players getting a _lot_ of mileage out of this

Sleet01 (approved):
LGTM!

PR #1764, opened 2025-03-12
Sheet preview in unit selection + draggable/undockable tabs

exeea:
Added functionality in the Unit Selection form to have a preview of the record sheet.
It support:
- zoom+pan
- multi-selection (multiple sheets visible)
- multi-page units are now visible *
- buttons for Print and Export to PDF, both supporting multi-selection.
- in the MainUI of each unit the main tabs are now draggable, this allows to drag the Preview tab and have unit editing while watching the Record Sheet update in realtime

\* additionally, small fix for advaero sheet that had variable not being replaced with the unit type (was displaying always %s).


pavelbraginskiy (changes requested):
Mostly all looks good, but it would be good to filter out units for which we can't present a record sheet, namely Turrets (Gun Emplacements) and Handheld Weapons. Right now, selecting a HHW causes an error message to popup, even if you're not viewing the sheet for it.

exeea:
Splitting the PR in smaller bites:
https://github.com/MegaMek/megameklab/pull/1765

PR #1802, opened 2025-03-30
Faster units loading, Editor panel as JPanel and dirty state of editors

exeea:
the scope of this PR:
- enhance loading speed of multiple units
- provide visual cues when an unit is in a dirty state (unsaved pending changes)
- remove the "not saved" warning if the unit is already saved

currently done:
- when loading an unit, we don't create an empty unit and then override it with our loaded unit. We directly load the unit without the intermediate.
- removed multiple redundant refreshes on unit loading
- loading multiple units will load only the first one as active tab, all other units will have their tab not activated
- lazy tab initialization for UI elements
- migrated MegaMekMainUI from a JFrame to a JPanel
- save and saveas are moved from MenuBar to MegaMekMainUI for better context
- implemented a listener for dirty state after changes in the UI panels
- Added a visual cue in the tabs to mark the ones that are dirty (bullet point replaces the X, VSCode style)
- Implemented a configurable warning when last tab is being closed, to prevent accidental application exit
- reorganized the flufftab, added image preview
- highlight in QuirksTab, selected quirks are now more prominent
- when closing the Configuration dialog, refresh all (instantly apply new RS rules, etc...)
- minor UI retouches

### DIFF
--- a/megameklab/src/megameklab/ui/dialog/settings/MiscSettingsPanel.java
+++ b/megameklab/src/megameklab/ui/dialog/settings/MiscSettingsPanel.java
@@ -23,9 +23,6 @@ import java.awt.FlowLayout;
 import java.io.File;
 import java.net.MalformedURLException;
 import java.net.URL;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.ResourceBundle;
 
 import javax.swing.*;
 import javax.swing.border.EmptyBorder;
@@ -55,6 +52,7 @@ public class MiscSettingsPanel extends JPanel {
 
     private final MMComboBox<MMLStartUp> startUpMMComboBox = new MMComboBox<>("StartUp", MMLStartUp.values());
     private final JCheckBox chkSummaryFormatTRO = new JCheckBox();
+    private final JCheckBox chkApplicationExitPrompt = new JCheckBox();
     private final JCheckBox chkSkipSavePrompts = new JCheckBox();
     private final JTextField txtUserDir = new JTextField(20);
     private final JSlider guiScale = new JSlider();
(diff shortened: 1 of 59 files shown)


pavelbraginskiy (changes requested):
My major annoyance with this feature is that focusing and un-focusing a field marks the unit as dirty even if I haven't changed anything. In fact, switching to and from the Structure/Armor tab marks the unit as dirty because switching to the Structure tab focuses the Chassis Name text box.

I'm not sure what the best way to resolve this is, but one possibility is to generate the mtf/blk data of the original unit when the editor is marked clean, and see if the current unit produces a different mtf/blk stream. Although, this might be slow, and the produced mtf/blk streams have a datestamp that make them nondeterministic. 

I also have a couple minor nitpicks.
- Newly created units should be marked dirty since they haven't ever been saved yet.
- Adjusting the Manual BV field doesn't mark dirty.

exeea:
- new units are now marked as dirty
- new smart listener for dirty changes, it detects everything and uses mtf/blk comparison
- it can also detect manual undo actions and bring back a dirty unit to clean state

megameklab/src/megameklab/ui/dialog/settings/MiscSettingsPanel.java
```diff
@@ -84,9 +82,6 @@ public class MiscSettingsPanel extends JPanel {
         mulOpenLine.add(Box.createHorizontalStrut(5));
         mulOpenLine.add(cbMulOpenBehaviour);
 
-        chkSummaryFormatTRO.setText(resources.getString("ConfigurationDialog.chkSummaryFormatTRO.text"));
```
exeea, line 87: This was duplicated code.

PR #1816, opened 2025-04-03
undo/redo/reload

exeea:
this PR implements:
- undo
- redo
- reload

under a new menu: Edit

this PR includes also:
- some rework of the createNewUnit() methods for the various units to properly assign the entity at the end of the init
- double click on unit selection screen, selects!
- few other bugfixes.


exeea:
> **megameklab/src/megameklab/ui/generalUnit/BasicInfoView.java:272**
> * [nitpick] The removal of the conditional check (previously 'if (en.getManualBV() >= 0)') means manualBV is always set, which may lead to unintended behavior if negative values should be ignored. Please confirm that this change is intentional.
> ```
> setManualBV(en.getManualBV());
> ```
> **megameklab/src/megameklab/ui/MegaMekLabMainUI.java:202**

Negative values are ignored automatically by setManualBV() and replaced by ""

megameklab/src/megameklab/ui/MegaMekLabMainUI.java
```diff
@@ -224,6 +402,7 @@ public String saveUnitToString(Entity entity, boolean includeGeneratorHeader) {
             }
             return unitAsString;
         } catch (Exception ex) {
+            logger.error("", ex);
```
rjhancock, line 405: Don't just blindly log the exception. Add some details about what was expected to occur and why it didn't. Helps with tracking more details in the logs.

megameklab/src/megameklab/ui/MegaMekLabTabbedUI.java
```diff
@@ -344,7 +359,7 @@ private JPopupMenu createLoadUnitPopupMenu() {
      * @return The currently selected MegaMekLabMainUI instance, which represents
      *         the active editor in the tabbed UI, or null if no tab is selected.
      */
-    public MegaMekLabMainUI currentEditor() {
+    public MegaMekLabMainUI getActiveEditor() {
         int selectedIndex = tabs.getSelectedIndex();
```
rjhancock, line 363: Deprecate the old public function and have it call the new one along with JavaDocs stating to use it.

Add the appropriate since tags as well as we know when to remove them from the code base.

megameklab/src/megameklab/ui/MegaMekLabMainUI.java
exeea: added an error description

rjhancock: Thank you. I realize I may have come across as rude, sorry for that.

megameklab/src/megameklab/ui/MegaMekLabTabbedUI.java
exeea: added deprecation

rjhancock (approved):

PR #1841, opened 2025-04-20
Security Fix for File open when passing via command line arguments

rjhancock:
This is a security issue recently found that would potentially allow opening of any file on the system and checking for its existence. This fixes that issue by confirming it ends with the extensions we want and doesn't include any known issues within the filename.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -36,6 +36,7 @@ jobs:
           languages: java-kotlin
           queries: security-and-quality
           build-mode: none
+          source-root: megameklab
           config: |
             paths:
               - megameklab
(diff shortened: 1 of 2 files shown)


pavelbraginskiy:
This is not a security vulnerability.

rjhancock:
Leave it open. It's marked for next dev cycle.

exeea:
I must say is not a vulnerability. 
The only thing MML can discover is the files an user running it can already access on their own machine.
This "attack" doesn't work from web because the single instance service server is bind on loopback.

rjhancock:
The issues this PR was meant to address was a potential path traversal ALONG WITH a Malformed XML reading of the MUL file that could cause an RCE due to the projects dependency on Stream and XML -> Java Objects.

Due to research provided by @Scoppio, with manual parsing of the XML, this is not an issue.

PR #1857, opened 2025-04-26
Fix some jank caused by printing from MM

pavelbraginskiy:
There's currently two problems with printing from MM, this fixes both.

- After the print queue dialog closes, the MML process wouldn't actually end. Force exit the process.
- If the dialog is minimized it could get "lost" without a parent frame registering with the OS, so we create one so that you can get back to the dialog by alt+tabbing to it or selecting it in the taskbar.


exeea:
I found that the root issue was the memory leak due to the dialogs never being disposed. I have done a variant of your fix here: #1862
It also fixes the fact that cancelling the "system print dialog" was closing the queue. Now it properly keep you in the queue until you complete the print or you cancel.

PR #1869, opened 2025-05-01
fixed structure gun pips on quad protomek

exeea:
This PR fixes the alignment of the structure pips on the gun


HammerGS (approved):
Looks good

PR #1871, opened 2025-05-03
heat view now displays also total dissipation and total heat

### LINKED ISSUE #568: Enhancement request: Total Heat
Description
It is very difficult to see the total heat when designing various unit types (WS,DS,SC, etc), it now requires me to take the data from the BV calculations and paste it into a spreadsheet. An enhancement would be show the total or maximum heat in the "Heat Sinks" section of the first tab.

exeea:
With this, all units that are tracking heat/heatsinks will now display, in the HeatSink group of the Structure Tab, also how much they are dissipating and the total heat generated by all the equipment.

Closes #568

image

### DIFF
--- a/megameklab/src/megameklab/ui/generalUnit/HeatSinkView.java
+++ b/megameklab/src/megameklab/ui/generalUnit/HeatSinkView.java
@@ -89,9 +89,13 @@ public class HeatSinkView extends BuildView implements ActionListener, ChangeLis
     private final JLabel lblPrototypeCount = new JLabel();
     private final JSpinner spnPrototypeCount = new JSpinner();
     private final JLabel lblCritFreeText = new JLabel();
-    private final JLabel lblCritFreeCount = new JLabel();
+    private final JTextField lblCritFreeCount = new JTextField();
     private final JLabel lblWeightFreeText = new JLabel();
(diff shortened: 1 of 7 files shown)


megameklab/src/megameklab/ui/generalUnit/HeatSinkView.java
```diff
@@ -395,4 +427,9 @@ private void reportChange() {
         }
     }
 
+    private int getTotalHeatGeneration(Entity entity) {
```
pavelbraginskiy, line 430: Can we reuse an existing method for calculating total heat instead of writing a new one?
So for example if we ever decide to double-count UACs in the RS heat calculation, the heat calculation would be improved here as well.

pavelbraginskiy (changes requested):

megameklab/src/megameklab/ui/generalUnit/HeatSinkView.java
exeea: We don't really have a method that does it, this is the same inline formula used inside the recordsheets to display the (optional) total heat at the end of the inventory list.
But yes, I will unify it under a single method in UnitUtil

pavelbraginskiy (approved):

PR #1876, opened 2025-05-07
Fixes display of dissipation

exeea:
Fixes the display of the dissipation capacity of a mech in the Heatsink group (Structure Tab).
Now will show only the dissipation (heatsink capacity)

extra: moved aero mod calculation to MM.


pavelbraginskiy:
> extra: moved aero mod calculation to MM.

Is there an MM PR that goes with this?

exeea:
It was already merged

PR #1883, opened 2025-05-08
Fixed large inventory text fitting

exeea:
This PR fixes the text fitting calculation to fill the inventory area in the record sheet.
It properly calculates the number or lines required and it uses a smarter approach for calculation of line spacing reduction and font reduction.


exeea:
before:
image

after:
image

pavelbraginskiy (approved):

pavelbraginskiy (changes requested):
image
image
image

PR #1889, opened 2025-05-17
Allow TAG on HHWs

pavelbraginskiy:
Because despite being a WeaponType MML considers TAG to be Equipment and not a Weapon.


Sleet01 (approved):
LGTM

PR #1891, opened 2025-05-18
fix: checks if OS is mac and dont load the html for the recent files in the menu bar

Scoppio:
# What it does?

On MegaMekLab the "Recent files" in the menu bar appear with a bunch of html tags written in raw text. Apparently when adapting the look and feel of the OS it also gains some of its quirks, mac in particular is not able or not willing to render HTML on the menu bar in the way we setup it right now.

This PR checks if the OS is mac or not, and if it is mac it won't load the html tags 

Before:
image

After:
Screenshot 2025-05-17 at 23 32 40

### DIFF
--- a/megameklab/src/megameklab/ui/util/OSUtil.java
+++ b/megameklab/src/megameklab/ui/util/OSUtil.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright (C) 2025 The MegaMek Team. All Rights Reserved.
+ *
+ * This file is part of MegaMek.
+ *
+ * MegaMek is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License (GPL),
+ * version 3 or (at your option) any later version,
+ * as published by the Free Software Foundation.
+ *
+ * MegaMek is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


exeea:
What about render it styled but without using HTML to keep the same look between OSes?

Scoppio:
> What about render it styled but without using HTML to keep the same look between OSes?

Just for reference to whoever see this PR later: - We tried this solution and it did not work. MacOS is annoying like that.

megameklab/src/megameklab/ui/util/OSUtil.java
exeea, line 1: Similar OS checks are used in:
PrintRecordSheet.java
SingleinstanceService.java
MekHQ.java

I think this helper is worth of megamek.common and then be used in all 4 places.

Sleet01 (approved):
LGTM

exeea (approved):
Approved!

megameklab/src/megameklab/ui/util/OSUtil.java
Scoppio: Agreed, after this release we do this move

PR #1892, opened 2025-05-18
recent units multiplatform version

exeea:
Variant of https://github.com/MegaMek/megameklab/pull/1891

Should work with OSX and Linux too while keeping the styled menuitems.

image


Scoppio:
Screenshot 2025-05-18 at 18 20 55

Dont actually work on Mac

PR #1905, opened 2025-05-25
feat: add checkstyle tied to test

Scoppio:
# What it does?

It does the same as this https://github.com/MegaMek/megamek/pull/7131


exeea (approved):
LGTM!

PR #1908, opened 2025-05-25
Legal: License update

rjhancock:
Needed updates to scripts and files for pending change.


.gitignore
```diff
@@ -38,3 +38,6 @@
 units.cache
 .mml_tmp
 *.iml
+
+/megameklab/docs/README.md
```
exeea, line 42: Why these ignore?

rjhancock: On all 3 it is being ignored in favor of the one in the root of the project that is copied to the docs folder.

exeea (approved):

PR #1914, opened 2025-05-28
Don't read the config if the setting is empty

exeea:
This PR adds a simple check to prevent read the content of a setting if is empty.


rjhancock (approved):
Sanity Check Complete.

PR #1916, opened 2025-05-28
Composite splash image

exeea:
This PR adds the same composite image approach for splash screen from MM to MML.


rjhancock:
As odd as it sounds, don't remove the unused assets. I'll be dealing with them in the data repo separately.

PR #1973, opened 2025-07-22
Don't include ammo twice on HHW RS

pavelbraginskiy:
Handheld Weapons have a special section on their record sheet dedicated to just ammo, therefore the ammo count does not need to also be included in the Weapons & Equipment Inventory section.

Before:
image

After:
image


Sleet01 (approved):
LGTM

PR #1974, opened 2025-07-23
Include ammo count in HHW ammo display

pavelbraginskiy:
Instead of needing to count pips to see how much ammo you have, also include a number. 
Single bin:
image
Multiple bins:
image


exeea (approved):
looks good!

PR #1986, opened 2025-07-31
Improvement: MHQ->MML Data Update

IllianiBird:
This PR just copies over data changes from mhq into mml so that the data merger can proceed without risk of losing mhq changes

This largely affects some recent faction data updates and faction images


Sleet01 (approved):
LGTM

PR #2013, opened 2025-09-11
Tweak spacing between Ht and Dmg columns

pavelbraginskiy:
Adds a little bit more space between the Ht and Dmg columns so that the Equipment inventory looks nicer.

Currently it can look a little awkward with two-digit heat: 
image

This moves the Ht column a _tiny_ bit to the left to create some visual separation, but not enough to overlap with the Loc column when it has split locations.
image


exeea (approved):
LGTM!

PR #2017, opened 2025-09-14
Add button to sort record sheets in print queue

### LINKED ISSUE #2015: [RFE] Sort files for printing
### Enhancement Type *

New Feature

### Brief Description of Enhancement *

When building large books of RS it would be nice to have some sort ability without manually doing it. RS Books have a standard sort order. 

Battle Armor
Vehicles
BattleMechs
ProtoMechs
Aerospace

Clan First, then Inner Sphere. 

Weights and Alphabetical for same weights. 

### Use Cases or Rationale

_No response_

### Attach Files

_No response_

### MegaMekLab Suite Version *

0.50.07-Nightly

### Final Verification

- [x] I confirm this request hasn't already been submitted (checked the tracker)
- [x] I've discussed or asked about this enhancement on MegaMek Discord
- [x] I’m opening this on the correct repo (MegaMek, MegaMekLab, or MekHQ)

pavelbraginskiy:
Closes #2015.

I considered adding support for configuration, but that adds a lot of ui work for something I don't think anybody will care about. Sorting first by unit type is the main thing I imagine anyone will want.

Alphabetical sort uses your configured unit print name, so the same name controlled by the *Inner Sphere (Clan)*/*Clan (Inner Sphere)* setting.

Page breaks are added to ensure units come out in the expected order.

The only thing that's a little wonky is that clantech variants of IS chassis will sort with the other clan units, so e.g. you won't get the Charger C together with the other Chargers. You'll have to put it in the right spot manually, but this isn't too hard and I don't think it's avoidable.

### DIFF
--- a/megameklab/src/megameklab/printing/PrintEntity.java
+++ b/megameklab/src/megameklab/printing/PrintEntity.java
@@ -905,8 +905,7 @@ public abstract class PrintEntity extends PrintRecordSheet {
     }
 
     private String entityName() {
-        return CConfig.getMekNameArrangement().printChassis(getEntity())
-              + (StringUtility.isNullOrBlank(getEntity().getModel()) ? "" : " " + getEntity().getModel());
+        return UnitUtil.getPrintName(getEntity());
     }
 
     protected boolean useAlternateArmorGrouping() {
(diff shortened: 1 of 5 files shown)


exeea (approved):
LGTM!

PR #2056, opened 2025-10-21
FIX: Issue 7400 - fixes missing faction availability for clan mechs with is names

GypsyFromTheSatelliteofLove:
fixes issue in MegaMek/megamek#7400

changed the 3 setUnit method calls to use entity.getFullChassis() instead of entity.getChassis()
also added a trim() at the AvailabilityPanel call to getModel from RatGenerator to make sure any blank models dont cause issues 

also the other classes are in Megamek repo
https://github.com/MegaMek/megamek/pull/7516#issue-3534874679
MegaMek/megamek#7516


psikomonkie (approved):
Nice!

PR #2064, opened 2025-10-31
fixed ground movement table

exeea:


Sleet01:
> > For changes to the record sheet, could we see visual examples of the change? I can figure out what the code does, but not how the result will differ visually.
> 
> It fixes this. We had it broken since months. (and the missing labels since more)
> 

Yep, that's broken :)

Sleet01 (approved):
LGTM.  Thanks for the comparison shot!

PR #2081, opened 2025-11-30
No BA JB on RS

pavelbraginskiy:
The BA Jump Booster (not to be confused with the BA Mechanical Jump Booster) does nothing but add 1 jump mp, there's no reason to include it in record sheets.


exeea (approved):
LGFM

PR #2129, opened 2026-02-07
Apply Vehicle Effectiveness Rules from TacOps:AR

bliebeg:
A menu option was added to toggle the Vehicle Effectiveness rules from TacOps:AR p.105 on Ground Combat Vehicle Record sheets. The changes include removing the Motive Checks from the 9 row of the Hit Location table, remove the Critical check from a Side roll of 8. The rules also remove the Attack Direction modifiers, and reduce the Vehicle Type modifiers by 1, minimum 0.

### DIFF
--- a/.github/workflows/checkstyle.yml
+++ b/.github/workflows/checkstyle.yml
@@ -1,14 +1,11 @@
-name: MegaMekLab CI with Gradle
+name: Checkstyle
 
 on:
   pull_request:
     branches: [ main ]
 
-env:
-  GRADLE_OPTS: "-Dscan.link.VCS=${{ github.event.pull_request.html_url }}"
-
 jobs:
-  pr_ci:
+  checkstyle:
     runs-on: ${{ matrix.os }}
 
     outputs:
@@ -19,7 +16,7 @@ jobs:
       matrix:
         os: [ ubuntu-latest ]
         java-distribution: [ temurin ]
-        java-version: [ 17, 21, 25 ]
+        java-version: [ 25 ]
       fail-fast: false
 
     steps:
@@ -28,11 +25,6 @@ jobs:
         with:
           repository: megamek/mm-data
           path: mm-data
-
-      - name: Checking out MegaMekLab
-        uses: actions/checkout@v6
-        with:
-          path: megameklab
 
       - name: Find the Right MegaMek Branch
         id: find_mm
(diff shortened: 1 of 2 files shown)


megameklab/resources/megameklab/resources/Dialogs.properties
```diff
@@ -117,6 +117,9 @@ ConfigurationDialog.chkTacOpsHeat.tooltip=Includes the expanded heat scale from
 ConfigurationDialog.chkFancyPips.text=Use pip shape to indicate special armor
 ConfigurationDialog.chkFancyPips.tooltip=<html>Circles by default, diamonds for hardened armor,<br>pentagons for \
   reflective, reactive, etc.</html>
+ConfigurationDialog.chkTacOpsVehicleEffectiveness.text=Reduced Vehicle Motive System hit chances
+ConfigurationDialog.chkTacOpsVehicleEffectiveness.tooltip=Implements Vehicle Effectiveness changes from TacOps:TO p\
```
pavelbraginskiy: I would actually switch around the information in the tooltip vs the text - if you call the checkbox something like "Use TacOps Vehicle Effectiveness table" it's immediately obvious to everyone who knows what that is what this options does, then you can use the tooltip to explain what that means to those who don't know what that is.

megameklab/src/megameklab/printing/RecordSheetOptions.java
```diff
@@ -427,4 +430,12 @@ public boolean fancyPips() {
     public void setFancyPips(boolean fancyPips) {
         this.fancyPips = fancyPips;
     }
+
+    public boolean TacOpsVehicleEffectiveness() { return tacOpsVehicleEffectiveness; }
+
+    public void setTacOpsVehicleEffectiveness(boolean tacOpsVehicleEffectiveness) {
+        this.tacOpsVehicleEffectiveness = true;
+        //this.tacOpsVehicleEffectiveness = tacOpsVehicleEffectiveness;
```
pavelbraginskiy: I think you forgot to uncomment this

megameklab/src/megameklab/ui/dialog/settings/ExportSettingsPanel.java
```diff
@@ -372,14 +377,17 @@ class ExportSettingsPanel extends JPanel {
         innerGridPanel.add(hitModPanel);
         innerGridPanel.add(intrinsicPhysicalsPanel);
         innerGridPanel.add(chkExtraPhysicals);
+        // innerGridPanel.add(chkTacOpsVehicleEffectiveness);
```
pavelbraginskiy: Please don't leave commented out code that you decided against

megameklab/resources/megameklab/resources/Dialogs.properties
bliebeg: Fixed.

megameklab/src/megameklab/printing/RecordSheetOptions.java
bliebeg: I did. I had that in there while I was troubleshooting. I ended up keeping the commented line, and removed the other. Line 437 doesn't actually take in the passed in parameter to do anything with. I tested with like 438 uncommented and 437 deleted and the functionality does not seem to be affected.

megameklab/src/megameklab/ui/dialog/settings/ExportSettingsPanel.java
bliebeg: Deleted.

Sleet01 (approved):
LGTM

PR #2199, opened 2026-05-18
Add option to hide rules level in record sheets

pavelbraginskiy:
Requires MegaMek/mm-data#359.

Adds on option to show the old rules level of the unit, defaulting to Off. 
When the value isn't shown, other elements of the sheet are moved around to save space and allow the Equipment Inventory to be larger.


Sleet01:
Dependent on MegaMek/mm-data#359 which needs its merge conflicts sorted.

Please update that PR and then ping me.

SJuliez (approved):
looks ok to me. Was the naming of "hidrole" and "hidRulesLevel" intentional, rather than "hideRole"?

PR #2292, opened 2026-08-12
Fix: Armor Tonnage on Armor Type Change

WellBattle6:
Fix armor tonnage on armor type change for CV, SV, BM, AS, and Advanced Aerospace

For CV, SV, BM, and AS, armor tonnage is now set to maximum tonnage on armor type change if the previous armor tonnage is greater than the maximum tonnage

For Advanced Aerospace, armor tonnage is no longer incorrectly set on armor type change

Note for Reviewer: MERGE AFTER https://github.com/MegaMek/megamek/pull/8689

### DIFF
--- a/megameklab/src/megameklab/ui/largeAero/WSStructureTab.java
+++ b/megameklab/src/megameklab/ui/largeAero/WSStructureTab.java
@@ -367,8 +367,6 @@ public class WSStructureTab extends ITab implements AdvancedAeroBuildListener, A
     public void armorTypeChanged(int at, int aTechLevel) {
         getJumpship().setArmorTechLevel(aTechLevel);
         getJumpship().setArmorType(at);
-        // recalculate tonnage
-        getJumpship().setArmorTonnage(getJumpship().getArmorWeight());
         panArmorAllocation.setFromEntity(getJumpship());
         panSummary.refresh();
         refresh.refreshStatus();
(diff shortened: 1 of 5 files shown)


megameklab/src/megameklab/ui/largeAero/WSStructureTab.java
```diff
@@ -367,8 +367,6 @@ public void heatSinkBaseCountChanged(int count) {
     public void armorTypeChanged(int at, int aTechLevel) {
         getJumpship().setArmorTechLevel(aTechLevel);
         getJumpship().setArmorType(at);
-        // recalculate tonnage
-        getJumpship().setArmorTonnage(getJumpship().getArmorWeight());
```
pavelbraginskiy, line 371: Can you explain removing this line?

It *looks* like this line being removed sets the tons of armor allocated equal the weight necessary for the new armor type to give the current number of pips, which is a perfectly reasonable thing to do here. 

What is it actually doing and why is it a problem?

WellBattle6: That deleted code sets the armor tonnage to fit the number of armor points BEFORE the armor type change, because the new armor points for the changed armor type haven't been applied to the entity till the user selects Auto-Allocate Armor or manually allocates per location. getArmorWeight first retrieves current/old entity armor points, then divides that by the new armorType's pointsPerTon to get tonnage.

If you look at the Summary box, without removing this, you'll notice the listed armor tonnage won't match the Armor Tonnage selector when you switch armor type. So when you 'upgrade' to an armor type with more points per tonnage, the ship is subtracting armor weight rather than adding more (unallocated) armor points, which is an anti-pattern to every other entity builder in MegaMekLab. (Likewise degrading armor type results in sudden added armor weight).

Screenshot 2026-08-16 at 6 51 15 PM
Screenshot 2026-08-16 at 6 51 21 PM
Notice how on changing from Standard Aerospace to Ferro-Carbide, the JumpShip tonnage in the Summary changes from 27.5 tons to 18.5 tons despite the Armor Tonnage selector remaining at 27.5.

pavelbraginskiy (approved):

PR #2296, opened 2026-08-16
GameRules and shield/charge printing

exeea:
This PR implements game rules selection and updated printing for shields and charge.

in core 2026:
Charge will use `Nx(TTM+1)` format
Shield will be backed inside the punch damage

in TW: remains the old format.


TenkawaBC:
Gate the shield damage modifier checks in IntrinsicPhysicalInventoryEntry and PhysicalAttacks behind an entity.hasShield() call

TenkawaBC (approved):
Validated the shield changes. Looks correct for both core and TW, and for when showing physical damage or not.

